PR #3, opened 2025-03-18
feat: add GitHub Actions workflows for Docker release and release drafter configuration

nookyo:
Introduce workflows to automate Docker releases and manage release drafting. The configuration includes version updates, artifact deployment, and conditional Docker builds based on the presence of a Dockerfile.

### DIFF
--- a/.github/workflows/build-on-commit.yaml
+++ b/.github/workflows/build-on-commit.yaml
@@ -1,7 +1,7 @@
 name: Build on Commit
 
 on:
- - push
+ workflow_dispatch:
 
 jobs:
     mvn:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,175 @@
+---
+
+# This GitHub Actions workflow is designed to be triggered when a release is marked as a full release.
+# The workflow performs the following tasks:
+# 1. Checks if the tag already exists.
+# 2. Updates the version in the pom.xml file.
+# 3. Commits the changes to the repository.
+# 4. Builds the project using Maven.
+# 5. Runs tests.
+# 6. Tags the commit with the release version.
+# 7. Deploys the artifact to the Maven repository.
+# 8. Builds and publishes a Docker image.
+# 9. Creates a GitHub release.
+
+# To make it work for your project, you need to adjust the pom.xml and add configuration file for GitHub release.
+# Please find detailed instructions:
+# https://github.com/Netcracker/qubership-workflow-hub?tab=readme-ov-file#maven-project-release-workflow
+
+name: Docker Release 
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        required: true
+        default: '1.0.0'
+        type: string
+        description: 'Release version (e.g., 1.0.0)'
+      java-version:
+        required: false
+        type: string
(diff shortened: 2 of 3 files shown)


.github/workflows/build-on-commit.yaml
```diff
@@ -1,7 +1,7 @@
 name: Build on Commit
 
 on:
- - push
+ workflow_dispatch:
```
lis0x90, line 4: Leave this intact. We want to build at least maven project to run unit test etc.
build docker or not - other story

.github/workflows/docker.yml
```diff
@@ -0,0 +1,175 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Update pom.xml
+        id: config
+        uses: netcracker/qubership-workflow-hub/actions/pom-updater@main
+        with:
+          path: //p:project/p:version
```
lis0x90, line 100: I doubt that is works as you expect.
May be better use default maven tools: mvn version:set ?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,175 @@
...
+    uses: netcracker/qubership-workflow-hub/.github/workflows/maven-publish.yml@main
+    with:
+      maven-command: "--batch-mode package -Dgpg.skip=true"
+      java-version: ${{ github.event.inputs.java-version }}
+      upload-artifact: true
+      artifact-id: ${{ needs.update-pom-version.outputs.artifact_id }}
+      server-id: ${{ github.event.inputs.mode }}
+    secrets:
+      maven-token: ${{ secrets.GITHUB_TOKEN }}
+      # maven-username: ${{ secrets.MAVEN_USER }}
```
lis0x90, line 119: Are these comments still need here?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,175 @@
...
+
+  tests:
+    needs: [mvn-package]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Run tests
+        run: echo "Running tests here"
```
lis0x90, line 132: What is this?

.github/workflows/docker.yml
nookyo: Yeah, it is just only for build.

.github/workflows/build-on-commit.yaml
nookyo: We have made an example for you that you can adapt as you need. This is an example of how to make a release using maved and docker publishing.

.github/workflows/docker.yml
nookyo: It is emty space for integration tests

.github/workflows/docker.yml
nookyo: Not quite, we need to get the source code in the image in which the version is replaced)

nookyo: The thing is that many projects do not use the <version> tag directly on pom.xml. The properties indicate the <revision> through which the version changes and we have to change file. If you do not need this step, you can safely turn it off.
example:

PR #7, opened 2025-03-27
feat: Feature/dev-docker-build

nookyo:
fix new docker dev build

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -58,7 +58,7 @@ jobs:
        needs: perform-version
        with:
          ref: ${{ github.event.ref }}
-         artifact-id: ${{ github.event.inputs.artifact-name }}
+         artifact-id: ${{ github.event.repository.name }}
          dry-run: ${{ inputs.dry-run }}        
          download-artifact: true
-         tags-override: ${{ needs.perform-version.outputs.metadata }}
+         tags: ${{ needs.perform-version.outputs.metadata }}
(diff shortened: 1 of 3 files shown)


.github/workflows/dev-build.yml
```diff
@@ -0,0 +1,79 @@
...
+      artifact-name:
+        description: 'Artifact name'
+        required: false
+        default: ''
+      tags:
+        description: 'Tags'
+        required: false
+        type: string
+        default: ''
+      java-version:
```
lis0x90, line 16: Externalize required java version to config in .github folder? 
It also will be useful for grand report, for example
Required maven version for buil is also main parameter for build

.github/workflows/dev-build.yml
```diff
@@ -0,0 +1,79 @@
...
+        default: false
+
+jobs:
+  dev-build:
+    uses: netcracker/qubership-workflow-hub/.github/workflows/maven-publish.yml@main
+    with:
+      maven-command: "--batch-mode package -Dgpg.skip=true -DskipTests=true"
+      java-version: ${{ github.event.inputs.java-version }}
+      upload-artifact: true
+      artifact-id: ${{ github.event.inputs.artifact-name || github.event.repository.name }}
```
lis0x90, line 34: artifact-id? What does it mean? Maven project has its own GAV coordinates

nookyo: ? it is our parametr? how it affected you?

.github/workflows/dev-build.yml
nookyo: ?

PR #90, opened 2025-07-29
Update OpenAPI.json

kichasov:


ArkuNC:
Changes should be done in the controller's annotations, and then applied with SwaggerOpenApiDocsTest.

PR #377, opened 2026-04-10
fix: controller code cleanup

TheHollowMonarch:


dbaas-operator/internal/client/aggregator_client.go
```diff
@@ -167,6 +155,16 @@ func (c *AggregatorClient) RegisterExternalDatabase(ctx context.Context, namespa
 	return nil
 }
 
+func decodeResponse(body []byte, context string) (*DeclarativeResponse, error) {
```
kichasov: The parameter name `context` shadows the standard `context` package. Consider renaming it to avoid potential confusion or collisions if a `context.Context` is ever added.

PR #415, opened 2026-04-27
feat: db policy migration to v1

TheHollowMonarch:


kichasov:
**`config/rbac/role.yaml` not regenerated**

The kubebuilder RBAC markers were added to `dbpolicy_controller.go`, but `make manifests` was not run — `config/rbac/role.yaml` is absent from the diff. Without it, the stable deployment won't have permissions for `dbpolicies` / `dbpolicies/status` and the controller will fail with `forbidden` on the first reconcile.

Please run `make manifests` and include the updated RBAC files.

kichasov:
**Helm ClusterRole not updated**

`helm-templates/dbaas-operator/templates/ClusterRole.yaml` (and `dev/k8s/operator.yaml`) are missing rules for `dbpolicies` and `dbpolicies/status`. The DbPolicy controller is now unconditional, but without these rules a Helm-deployed operator will get `forbidden` on every reconcile.

kichasov:
**`config/kustomization.yaml` — is the DbPolicy CRD included?**

`config/crd/bases/dbaas.netcracker.com_dbpolicies.yaml` was added, but it's not clear whether `config/kustomization.yaml` references it. If not, `kustomize build config/` won't apply the CRD and the stable deployment will be missing it entirely.

PR #424, opened 2026-05-05
feat: adding a watcher for changes in secrets

TheHollowMonarch:


dbaas-operator/internal/controller/externaldatabase_controller.go
```diff
@@ -251,6 +255,19 @@ func (r *ExternalDatabaseReconciler) applySecretCredentials(
 // Reconcile returns an error (BackingOff phase).  Pass
 // ctrlcontroller.Options{} to keep the controller-runtime defaults.
 func (r *ExternalDatabaseReconciler) SetupWithManager(mgr ctrl.Manager, opts ctrlcontroller.Options) error {
+	// Register the field index before building the controller.
+	// controller-runtime calls indexSecretNames for every ExternalDatabase to
```
kichasov, line 255: The index key is the secret name only, without namespace:

```go
return []string{secretName}
```

```go
client.MatchingFields{secretNamesIndex: obj.GetName()}
```

If namespace A has a secret named `my-creds` and namespace B has an `ExternalDatabase` also referencing a secret named `my-creds`, a change in A's secret will trigger a spurious reconcile of B's `ExternalDatabase`. Not a correctness issue (B's reconcile reads its own namespace's secret), but unnecessary load and potentially confusing behaviour.

**Fix:** include namespace in the key:

```go
// indexSecretNames
return []string{obj.GetNamespace() + "/" + secretName}

// enqueueForSecret
client.MatchingFields{secretNamesIndex: obj.GetNamespace() + "/" + obj.GetName()}
```

dbaas-operator/config/rbac/role.yaml
```diff
@@ -10,6 +10,8 @@ rules:
   - secrets
   verbs:
   - get
+  - list
```
kichasov, line 13: `config/rbac/role.yaml` is updated, but the same change is missing in three other places:

- `config-dev/rbac/role.yaml` — not in the diff
- `dev/k8s/operator.yaml` — inline `ClusterRole` not updated
- `helm-templates/dbaas-operator/templates/ClusterRole.yaml` — not updated

Without the Helm template update, a production deployment will fail with `forbidden` on secrets list/watch.

Also worth noting: cluster-wide `list` + `watch` on `secrets` is a significant RBAC expansion — the operator will now receive events for every Secret in every namespace. Consider whether this scope is acceptable or if it should be bounded (e.g. via a namespace-scoped `Role` instead of `ClusterRole` for secrets).

kichasov:
**`enqueueForSecret` — missing test coverage**

The PR adds tests for `indexSecretNames` (good), but the most critical piece — `enqueueForSecret` — has no test coverage.

What's missing: an integration test that verifies a Secret change actually enqueues the correct `ExternalDatabase` objects. Something like:

1. Create an `ExternalDatabase` referencing `secret-a`
2. Trigger a reconcile so the field index is populated
3. Simulate a Secret event for `secret-a`
4. Assert that a reconcile request for the `ExternalDatabase` was produced

Without this, the wiring between the Watch and the reconcile queue is only verified manually.

dbaas-operator/internal/controller/externaldatabase_controller.go
```diff
@@ -278,3 +301,40 @@ func (r *ExternalDatabaseReconciler) enqueueForBinding(ctx context.Context, obj
...
+// indexSecretNames is the indexer function registered at startup.
+// controller-runtime calls it for every ExternalDatabase to populate and
+// maintain the reverse map: secretName → []ExternalDatabase.
+func indexSecretNames(obj client.Object) []string {
+	edb, ok := obj.(*dbaasv1.ExternalDatabase)
+	if !ok {
+		return nil
+	}
+	var names []string
+	for _, cp := range edb.Spec.ConnectionProperties {
```
kichasov, line 314: The Secret watch has no predicate, so any event — including label/annotation changes that don't affect credentials — will trigger a reconcile. Consider adding `predicate.ResourceVersionChangedPredicate{}`, which fires only when the object actually changes. Secrets have no `generation` field, so `GenerationChangedPredicate` does not apply here, but `ResourceVersionChangedPredicate` achieves the same effect.

```go
WatchesMetadata(&corev1.Secret{},
    handler.EnqueueRequestsFromMapFunc(r.enqueueForSecret),
    builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
)
```

dbaas-operator/internal/controller/externaldatabase_controller.go
```diff
@@ -278,3 +301,40 @@ func (r *ExternalDatabaseReconciler) enqueueForBinding(ctx context.Context, obj
...
+	return names
+}
+
+// enqueueForSecret maps a Secret event to reconcile requests for ExternalDatabases
+// that reference it. Uses the field index for an O(1) lookup — no full list scan.
+func (r *ExternalDatabaseReconciler) enqueueForSecret(ctx context.Context, obj client.Object) []reconcile.Request {
+	list := &dbaasv1.ExternalDatabaseList{}
+	if err := r.List(ctx, list,
+		client.InNamespace(obj.GetNamespace()),
+		// MatchingFields resolves against the in-memory index — no scan of unrelated EDBs.
```
kichasov: The indexer is registered exclusively for `ExternalDatabase`, so a non-EDB object here is a bug (wrong registration), not a normal case. Silently returning `nil` would hide the problem. A warning log would make misconfigurations visible:

```go
if !ok {
    log.Warnf("indexSecretNames: unexpected object type %T", obj)
    return nil
}
```

dbaas-operator/internal/controller/externaldatabase_controller.go
```diff
@@ -251,6 +255,19 @@ func (r *ExternalDatabaseReconciler) applySecretCredentials(
...
 func (r *ExternalDatabaseReconciler) SetupWithManager(mgr ctrl.Manager, opts ctrlcontroller.Options) error {
+	// Register the field index before building the controller.
+	// controller-runtime calls indexSecretNames for every ExternalDatabase to
+	// build the reverse map (secretName → EDBs), then keeps it up to date
+	// automatically as EDBs are created, updated, or deleted.
+	if err := mgr.GetFieldIndexer().IndexField(
+		context.Background(),
+		&dbaasv1.ExternalDatabase{},
+		secretNamesIndex,
+		indexSecretNames,
```
kichasov, line 262: `secretNamesIndex` is defined inside `externaldatabase_controller.go`, but index key constants are better placed in `conditions.go` (alongside other shared controller constants) or a dedicated `indexes.go` file. If other controllers add their own field indexes in the future, having them scattered across controller files makes it harder to avoid naming collisions and to get a quick overview of all registered indexes.

PR #474, opened 2026-05-28
feat: DB balancing rules

TheHollowMonarch:
## Namespace Rule order has been made as mandatory

order is mandatory for namespace balancing rules because it controls rule priority in the aggregator. When multiple namespace balancing rules can match the same database type, the higher order value has priority.

The aggregator API accepts namespace rule updates where order may be omitted, but omitting it can preserve or rely on existing aggregator-side state. That is risky for operator-managed CRs because the desired state in Kubernetes would no longer fully describe the rule being applied.


dbaas-operator/api/v1/balancingrule_types.go
```diff
@@ -0,0 +1,331 @@
...
+	Order int64 `json:"order"`
+}
+
+// DbNamespaceBalancingRuleStatus defines the observed state of DbNamespaceBalancingRule.
+type DbNamespaceBalancingRuleStatus struct {
+	OperatorStatus `json:",inline"`
+
+	// appliedRules are the namespace rule entries last successfully applied
+	// to dbaas-aggregator. They are informational only; namespace rule cleanup
+	// is not performed because the aggregator API has no individual delete endpoint.
```
kichasov: **Namespace rules can't be cleaned up via the current API, and the silent drift is risky.**

Confirmed against the aggregator code:
- `BalancingRulesControllerV3` exposes only `PUT /balancing/rules/{ruleName}` (`saveOnNamespace`, upsert by name) — there is no DELETE for namespace rules.
- Sending an "empty" rule to neutralize it is **not** viable: an empty `config` NPEs in `getPhysicalDatabaseIdentifier`, and `{perNamespace:{}}` stores a `null` `physicalDatabaseIdentifier`. Unlike the microservice path (which returns `null` → default on empty rules), `getDatabaseForNamespace` picks the max-`order` rule and throws `UnregisteredPhysicalDatabaseException` on a missing phydbid — so a "neutralized" rule actively **breaks provisioning** for that `(namespace, type)`.

So today, removing an item from `spec.rules` (or deleting the CR — there's no finalizer) leaves the rule applied while the CR reports success, which breaks the "K8s fully describes the applied state" contract this PR is built on.

The data layer already supports deletion (`BalancingRulesService.removeRulesByNamespace`, `findByName`), so the clean fix is a new aggregator endpoint `DELETE /api/v3/dbaas/{namespace}/physical_databases/balancing/rules/{ruleName}`, after which this controller can clean up symmetric to permanent rules (finalizer + diff of `status.appliedRules`).

Until that endpoint exists, please make the limitation **explicit instead of silent**: emit a `Warning` event + a status condition listing the orphaned rule name(s) when an applied rule disappears from `spec.rules` or the CR is deleted.

dbaas-operator/internal/controller/balancingrule_reconcile_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+	dbaasv1 "github.com/netcracker/qubership-dbaas/dbaas-operator/api/v1"
+	aggregatorclient "github.com/netcracker/qubership-dbaas/dbaas-operator/internal/client"
+	"github.com/netcracker/qubership-dbaas/dbaas-operator/internal/ownership"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
```
kichasov: **Controller tests here use `testing.T` + fake client, which diverges from the project's test convention.** `CLAUDE.md` mandates envtest + Ginkgo/Gomega for controller tests, and every sibling `*_controller_test.go` (databasesecret, externaldatabase, dbpolicy, namespacebinding) follows that. Two concerns:

1. Stylistic inconsistency within the `controller` package.
2. The fake client bypasses the API server, so the CRD-level schema (the now-mandatory `order`, `MinItems`, the label `Pattern`, singleton name) is never actually exercised — only the controller's redundant Go validation is.

Could you port these (`balancingrule_{payload,reconcile,validation}_test.go`) to the Ginkgo/envtest suite, or at least add envtest coverage that proves the CRD schema rejects bad specs? The `internal/client` tests staying on `testing.T` is fine — that's outside the controller package.

dbaas-operator/internal/controller/balancingrule_controller.go
```diff
@@ -0,0 +1,877 @@
...
+	}
+	if r.Client != nil {
+		if msg, err := r.validateNamespaceRuleGlobalConflicts(ctx, rule, names, typeOrders); msg != "" || err != nil {
+			return msg, err
+		}
+	}
+	return "", nil
+}
+
+func (r *BalancingRuleReconciler) validateNamespaceRuleGlobalConflicts(
```
kichasov, line 692: **This global uniqueness check is best-effort (racy) and partly duplicates server-side enforcement.** It `List()`s every `DbNamespaceBalancingRule` cluster-wide on each reconcile and checks `name`/`(type,order)` in memory, so two CRs reconciling concurrently can both pass and then apply conflicting `order`s (TOCTOU), and it's O(total rules) per reconcile.

The aggregator already enforces `(order, type)` uniqueness authoritatively: `saveOnNamespace` throws `BalancingRuleConflictException` (409), which this controller already maps to `InvalidConfiguration` via `handleAggregatorError`. So the `order` part of this pre-check is largely redundant — the 409 is the real guard.

The **name** uniqueness check, however, is valuable: namespace rules are keyed globally by name in the aggregator (`findByName`), so two CRs sharing a `ruleName` would silently clobber each other without it. Suggestion: keep the name check, but add a comment that it's best-effort and that `(order, type)` conflicts are ultimately enforced server-side (409) — and rely on that rather than treating the pre-check as a guarantee.

dbaas-operator/internal/controller/balancingrule_controller.go
```diff
@@ -0,0 +1,877 @@
...
+
+func namespaceRuleTriggerKey(namespace, name string) string {
+	return "namespace/" + namespace + "/" + name
+}
+
+func permanentRuleTriggerKey(namespace, name string) string {
+	return "permanent/" + namespace + "/" + name
+}
+
+func containsString(values []string, value string) bool {
```
kichasov: **Minor lint cleanups** (the repo enables `modernize` + `revive`, so CI may flag these):
- `containsString` duplicates the stdlib `slices.Contains` — consider dropping it.
- `lower` is a trivial `strings.ToLower` wrapper; inlining it reads about the same.
- `validateMicroserviceRule(ctx, …)` and `validatePermanentRule(ctx, …)` don't use `ctx` (unlike `validateNamespaceRule`) — `revive`'s unused-parameter may complain.

Worth a quick `make lint` pass.

dbaas-operator/internal/controller/balancingrule_controller.go
```diff
@@ -0,0 +1,877 @@
...
+		For(&dbaasv1.DbPermanentBalancingRule{},
+			builder.WithPredicates(generationOrLifecycleChangedPredicate())).
+		Watches(&dbaasv1.NamespaceBinding{},
+			handler.EnqueueRequestsFromMapFunc(r.enqueuePermanentRulesForBinding)).
+		WithOptions(opts).
+		Named("dbpermanentbalancingrule").
+		Complete(reconcile.Func(r.ReconcilePermanent))
+}
+
+func (r *BalancingRuleReconciler) reconcileMicroserviceDelete(
```
kichasov, line 410: **The deletion paths and the `WaitingForDependency` branch aren't covered.** The cleanup helpers have unit tests, but there's no end-to-end test for:
- `reconcileMicroserviceDelete` / `reconcilePermanentDelete` — finalizer removal + cleanup invocation when the CR has a `DeletionTimestamp`;
- the permanent-rule path where a target namespace isn't owned yet → `PhaseWaitingForDependency` + requeue.

These are the highest-risk flows (a missed finalizer removal strands the CR in `Terminating`; a skipped cleanup orphans rules). There's a test for "finalizer added before apply" but none for "finalizer removed + cleanup called on delete". Could you add coverage for both?

kichasov:
**Docs + PR description.** This PR introduces three user-facing CRDs but doesn't update `docs/howto/DBaaS Operator.md` (the secret-rotation PR set the precedent of documenting new CRs — spec fields, lifecycle, phases, RBAC). Please add a section covering at least: the singleton naming requirement, the per-CR lifecycle (incl. the namespace-rule cleanup limitation noted separately), the permanent rule living in the operator namespace while targeting other namespaces, and the namespace-ownership prerequisite. Also, the PR body on GitHub is currently empty — a short summary of scope and the "order is mandatory" rationale would help reviewers and future readers.

dbaas-operator/internal/controller/balancingrule_controller.go
```diff
@@ -0,0 +1,951 @@
...
+		err = r.Aggregator.ApplyNamespaceBalancingRule(ctx, rule.Namespace, item.Name, namespaceRequestFromSpecItem(item))
+		recordAggregatorCall(controllerBR, operationApplyNamespaceRule, aggStart, err)
+		if err != nil {
+			log.ErrorC(ctx, "failed to apply DbNamespaceBalancingRule to dbaas-aggregator: %v", err)
+			return handleAggregatorError(&rule.Status.Phase, &rule.Status.Conditions, rule.Generation, r.Recorder, rule, err, requestID)
+		}
+	}
+
+	markSucceeded(&rule.Status.Phase, &rule.Status.Conditions, rule.Generation, EventReasonBalancingRuleApplied)
+	rule.Status.AppliedRules = appliedNamespaceRulesFromSpec(rule.Spec.Rules)
```
kichasov: **[D1] Namespace rules apply per-item but `status.AppliedRules` is written only on full success — a partial apply failure orphans already-applied rules.**

Unlike `ReconcileMicroservice`/`ReconcilePermanent` (single bulk call), namespace rules are applied one-by-one in this loop, while `status.AppliedRules` is set only after the whole loop succeeds (line 239). If `apply` of rule *A* succeeds and rule *B* then fails (e.g. an aggregator `409` from another CR's `order`, or a `5xx`), `handleAggregatorError` returns with `AppliedRules` still holding the *pre-apply* value (empty on first apply).

Consequences — both break the very "K8s fully describes the applied state" contract this PR restores:
- **Delete:** `reconcileNamespaceDelete` iterates `AppliedRules`, so rule *A* (applied in the aggregator, absent from `AppliedRules`) is never deleted → finalizer is removed, CR disappears, rule *A* is orphaned forever.
- **Spec narrowing:** `cleanupSupersededNamespaceRules` diffs against `AppliedRules`, so rule *A* is never cleaned up.

This is especially likely when *B* maps to `InvalidConfiguration` (4xx, no auto-requeue): the user sees the error and may just delete the CR → orphan.

Fix: accumulate `AppliedRules` incrementally — append each entry right after its `ApplyNamespaceBalancingRule` succeeds, and persist it before returning on error. `patchStatusOnExit` already persists status on `InvalidConfiguration`, so the partially-applied set survives and is cleaned up correctly on later delete/narrowing.

dbaas-operator/internal/controller/balancingrule_controller.go
```diff
@@ -0,0 +1,951 @@
...
+		nameKey := strings.ToLower(item.Name)
+		if _, ok := names[nameKey]; ok {
+			return fmt.Sprintf("spec.rules contains duplicate name %q", item.Name), nil
+		}
+		names[nameKey] = struct{}{}
+		if strings.TrimSpace(item.Type) == "" {
+			return fmt.Sprintf("spec.rules[%d].type must not be blank", i), nil
+		}
+		if strings.TrimSpace(item.PhysicalDatabaseID) == "" {
+			return fmt.Sprintf("spec.rules[%d].physicalDatabaseId must not be blank", i), nil
```
kichasov, line 673: **[D2] `physicalDatabaseId` / `label` is validated for shape only, not existence — a rule pointing at a non-existent physical DB still reports `Succeeded`.**

Validation here only checks `physicalDatabaseId` is non-blank (and microservice `label` matches `key=value`). Neither the operator nor the aggregator's apply path (`saveOnNamespace` → `getPhysicalDatabaseIdentifier`) verifies the target physical DB is registered (or that a microservice `label` resolves to exactly one DB). So a typo'd id or an unmatched label is accepted and the CR reports `status: Succeeded`; the failure only surfaces later, at provisioning time, as `UnregisteredPhysicalDatabaseException` — possibly in a different namespace far from whoever authored the rule.

Consider either validating resolvability at apply time (aggregator returns 4xx for an unknown phydbid / unresolvable label → mapped to `InvalidConfiguration`), or documenting that `Succeeded` means "rule stored", not "target exists", so the status isn't read as a guarantee.

dbaas-operator/cmd/main.go
```diff
@@ -197,12 +197,35 @@ func main() {
...
+	namespaceRuleChecker := ownership.NewKindChecker(
+		mgr.GetClient(),
+		func() *dbaasv1.DbNamespaceBalancingRuleList { return &dbaasv1.DbNamespaceBalancingRuleList{} },
+		func(l *dbaasv1.DbNamespaceBalancingRuleList) int { return len(l.Items) },
+	)
+	permanentRuleChecker := ownership.NewKindChecker(
+		mgr.GetClient(),
+		func() *dbaasv1.DbPermanentBalancingRuleList { return &dbaasv1.DbPermanentBalancingRuleList{} },
+		func(l *dbaasv1.DbPermanentBalancingRuleList) int { return len(l.Items) },
+	)
```
kichasov: **[P1] `permanentRuleChecker` can't block NamespaceBinding deletion for a permanent rule's *target* namespace.**

`KindChecker.HasBlockingResources(ns)` lists the kind with `client.InNamespace(ns)` (checker.go:58), so it only blocks release when a CR of that kind physically lives in the namespace being released. But a `DbPermanentBalancingRule` must live in the **operator** namespace (`validatePermanentRule`, balancingrule_controller.go:653) and targets business namespaces only via `spec.namespaces`. So for a business namespace this checker always returns 0 and never blocks — unlike the microservice/namespace rule checkers, whose CRs are namespaced singletons in the business namespace itself.

Result: a `NamespaceBinding` can be deleted while a permanent rule still targets that namespace. The permanent reconcile then only goes to `WaitingForDependency` (line 299) without cleaning up; the aggregator-side permanent rule survives the release and can outlive a rebind of the namespace to a different operator instance (stale cross-operator designation).

Fix: either make the binding block on permanent rules by scanning `DbPermanentBalancingRule.spec.namespaces` / `status.appliedRules` cluster-wide (not `InNamespace`), or clean up the permanent entries for that namespace before the binding is released.

docs/howto/DBaaS Operator.md
```diff
@@ -1321,6 +1382,246 @@ kubectl get dbdd my-app-db -n my-namespace -o jsonpath='{.status.lastRequestId}'
...
+
+| Kind | On create/update | On item removal from `spec.rules` | On CR deletion |
+|------|------------------|------------------------------------|----------------|
+| `DbMicroserviceBalancingRule` | Adds a finalizer, applies the full desired list, stores applied `type + microservices`. | Sends cleanup for removed applied `type + microservices` by applying an empty rule set for those entries, then applies the new desired list. | Finalizer cleans up all applied microservice entries before Kubernetes removes the CR. |
+| `DbNamespaceBalancingRule` | Applies each desired namespace rule by name and stores applied entries for observability. | The aggregator currently has no individual namespace-rule delete endpoint. The controller emits a Warning event, sets `Ready=False` with reason `NamespaceRuleCleanupUnsupported`, and keeps the orphaned applied rule name in status. | No finalizer is added. Deleting the CR does not delete aggregator-side namespace rules until a delete API exists or external cleanup handles them. |
```
kichasov: **[P3] Stale runbook: this row describes the pre-finalizer namespace-rule behavior that this PR replaces.**

This row still states namespace rules have "no individual namespace-rule delete endpoint", report `NamespaceRuleCleanupUnsupported`, and that "No finalizer is added. Deleting the CR does not delete aggregator-side namespace rules". But this PR adds the `DELETE .../balancing/rules/{ruleName}` endpoint, `DbNamespaceBalancingRuleFinalizer`, and `reconcileNamespaceDelete` / `cleanupSupersededNamespaceRules` — so the CR now deletes aggregator-side rules on item removal and CR deletion, symmetric to permanent rules. The "intentionally explicit in status" sentence (line 1516) and the `NamespaceRuleCleanupUnsupported` reason row (line 1542) are outdated too. Please update the namespace row to match the new finalizer + DELETE cleanup.

TheHollowMonarch: Updated the doc

dbaas-operator/cmd/main.go
TheHollowMonarch: Making sure that a namespace binding can't be deleted if a permanent rule CR has that namespace in it.

dbaas-operator/internal/controller/balancingrule_controller.go
TheHollowMonarch: Documented that succeeded indicates that the rule got accepted by the aggregator

dbaas-operator/internal/controller/balancingrule_controller.go
TheHollowMonarch: Applied the fix

dbaas-operator/internal/controller/balancingrule_controller.go
```diff
@@ -0,0 +1,961 @@
...
+	}
+
+	appliedRules := make([]dbaasv1.DbNamespaceBalancingRuleAppliedRule, 0, len(rule.Spec.Rules))
+
+	for _, item := range rule.Spec.Rules {
+		aggStart := time.Now()
+		err = r.Aggregator.ApplyNamespaceBalancingRule(ctx, rule.Namespace, item.Name, namespaceRequestFromSpecItem(item))
+		recordAggregatorCall(controllerBR, operationApplyNamespaceRule, aggStart, err)
+		if err != nil {
+			rule.Status.AppliedRules = appliedRules
```
kichasov: **[Follow-up on the D1 fix] The incremental `AppliedRules` change is incomplete for update/retry — it starts empty and overwrites, dropping previously-applied rules that still exist aggregator-side.**

`appliedRules` starts empty each reconcile (line 228) and, on an apply error, overwrites `status.appliedRules` with only *this pass's* successes (line 235). Scenario: an earlier reconcile succeeded with `[A,B]` (aggregator has A,B). A later reconcile applies A, then B fails with a `4xx` → `InvalidConfiguration` — which `patchStatusOnExit` *does* persist — so `status.appliedRules` becomes `[A]` while B still exists in the aggregator. If the CR is then deleted, `reconcileNamespaceDelete` / `cleanupSupersededNamespaceRules` iterate only `status.appliedRules` (`[A]`) → **B is orphaned**.

(A `5xx`→`BackingOff` isn't persisted by the status predicate, so the leak is specific to the `4xx`/`InvalidConfiguration` path — but the overwrite is conceptually wrong regardless: status must reflect everything applied aggregator-side, not just the current pass.)

Suggest seeding `appliedRules` from the existing `status.appliedRules` and merging: keep a previously-applied entry until it has been successfully *removed* or *replaced* aggregator-side, rather than dropping it because of an unrelated mid-list failure on another rule.

---

**✅ Resolved in 60577d4** (pushed to `feat/dbIDRules`).

`status.appliedRules` now always reflects what the aggregator holds:

- `cleanupSupersededNamespaceRules` prunes each *successfully* deleted rule from `status.appliedRules`, and on a failed delete keeps the not-yet-deleted (still-live) entries recorded so the next reconcile retries them instead of orphaning them.
- The apply loop seeds the applied set from the (pruned) `status.appliedRules` and **upserts an entry only after a successful apply**, so a mid-list apply failure on one rule no longer drops the previously-applied ones.

Tests added:
- envtest: status holds `[mongo, cassandra]`; an updated spec prepends a new rule whose apply fails `409` → only the `payments-redis` `PUT` is attempted, phase is `InvalidConfiguration`, and `status.appliedRules` still equals `[mongo, cassandra]` (previously it would have been emptied → orphan).
- unit: `cleanupSupersededNamespaceRules` prunes the deleted rule from `status.appliedRules`.

kichasov:
**Round 3 — all round-1/round-2 findings addressed; the D1 update/retry follow-up is now resolved.**

Re-reviewed after `a6f2571`. The round-2 findings are resolved:

- **D2** (designator shape-only validation): documented on every designator field — a `Succeeded` status means dbaas-aggregator accepted/stored the rule, not that the `physicalDatabaseId` / `label` resolves to a registered physical DB. ✅
- **P1** (binding can't block on permanent rules): the useless `NewKindChecker(DbPermanentBalancingRuleList)` (which listed `InNamespace(business)` and never matched) is replaced by `NewPermanentBalancingRuleChecker(client, cloudNamespace)`, which lists permanent rules in the operator namespace and blocks release when the business namespace appears in `spec.Rules[].Namespaces` or `status.AppliedRules[].Namespaces`. + a `resolver_test.go` test. ✅
- **P3** (stale runbook): the namespace row, the blue-green finalizer note, and the `NamespaceRuleCleanupUnsupported` reason were all updated to the finalizer + DELETE behavior. ✅

Round-1 findings (Ginkgo port, best-effort uniqueness comment, lint, delete/finalizer/waiting coverage) remain addressed.

**✅ D1 (incl. the update/retry follow-up) — resolved in 60577d4.** The first-apply orphan was already fixed; the remaining update/retry orphan is now closed too. `status.appliedRules` always reflects what the aggregator holds: `cleanupSupersededNamespaceRules` prunes only *successfully* deleted rules (and on a failed delete keeps the not-yet-deleted, still-live entries recorded for retry), and the apply loop seeds the applied set from the pruned status and **upserts an entry only after a successful apply** — so a mid-list `4xx`→`InvalidConfiguration` no longer drops previously-applied rules. Covered by an envtest (status keeps `[mongo, cassandra]` when a prepended new rule fails `409`) and a strengthened `cleanupSupersededNamespaceRules` unit test. Details: https://github.com/Netcracker/qubership-dbaas/pull/474#discussion_r3381770566

No blockers from my side now. (Unrelated note: the `feat/dbIDRules` branch already fails `make lint` on ~11 pre-existing issues — e.g. the unused `appliedNamespaceRulesFromSpec`, several `unparam` — worth a separate cleanup pass.)

PR #493, opened 2026-06-05
feat: DBaaS password rotation and callback to operator

kichasov:


dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/OperatorEventDispatcher.java
```diff
@@ -0,0 +1,123 @@
...
+    ) {
+        this.rotationProperty = rotationProperty;
+        this.operatorEventRepository = operatorEventRepository;
+        this.operatorEventMetrics = operatorEventMetrics;
+        this.operatorWebhook = operatorWebhook;
+    }
+
+    @Scheduled(every = "${dbaas.operator.notification.dispatch-interval:5s}", concurrentExecution = SKIP)
+    @Transactional
+    public void dispatch() {
```
kichasov, line 48: **`dispatch()` performs network I/O inside the DB transaction holding `FOR UPDATE` row locks.**

`dispatch()` is `@Transactional`, and `claimPendingBatch()` takes `FOR UPDATE SKIP LOCKED` row locks on up to `BATCH_SIZE` rows. The per-event `operatorWebhook.notify()` HTTP calls then run inside that same transaction, so the row locks and the pooled DB connection are held for the entire duration of all HTTP calls in the batch. If the operator is slow or unreachable, that is up to `BATCH_SIZE × timeout` per transaction (worsened by the hardcoded 3-minute client timeout and the `@Retry(maxRetries=5)` on the REST client).

Risks: DB connection-pool exhaustion (a connection is pinned for the whole network wait), long-running transactions (table bloat / autovacuum starvation), and prolonged lock holding.

Suggest splitting the phases so network I/O happens **outside** the transaction: claim in a short transaction, send over HTTP unlocked, then persist each event's status in its own short transaction. The outbox stays multi-replica-safe via `SKIP LOCKED` while keeping each transaction in the millisecond range.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/OperatorWebhookClientFactory.java
```diff
@@ -0,0 +1,43 @@
...
+    @ApplicationScoped
+    public OperatorWebhook createOperatorWebhook() {
+        KubernetesTokenAuthFilter kubernetesTokenAuthFilter
+                = new KubernetesTokenAuthFilter(() -> KubernetesAudienceToken.getToken(rotationNotificationProperty.jwtAudience()));
+
+        OperatorWebhookRestClient restClient = RestClientBuilder.newBuilder().baseUri(URI.create(rotationNotificationProperty.callbackUrl()))
+                .register(kubernetesTokenAuthFilter)
+                .register(new DbaasAdapterRestClientLoggingFilter())
+                .connectTimeout(3, TimeUnit.MINUTES)
+                .readTimeout(3, TimeUnit.MINUTES)
```
kichasov: **Hardcoded 3-minute timeouts ignore the configured `connect-timeout`/`read-timeout` (`PT30S`).**

`RotationNotificationProperty` exposes `connectTimeout()` / `readTimeout()`, and `application.properties` defines `dbaas.operator.notification.connect-timeout`/`read-timeout` (default `PT30S`) — but this factory hardcodes `3, TimeUnit.MINUTES` for both, so those config properties are dead and the effective timeout is 3 minutes rather than the documented 30 seconds.

Combined with the `@Retry(maxRetries=5)` on the REST client and the in-transaction dispatch loop, a stuck operator can block a single `notify()` for up to `5 × 3min` while holding DB row locks.

Use the configured values, e.g.:
```java
.connectTimeout(rotationNotificationProperty.connectTimeout().toMillis(), TimeUnit.MILLISECONDS)
.readTimeout(rotationNotificationProperty.readTimeout().toMillis(), TimeUnit.MILLISECONDS)
```

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/rest/OperatorWebhookRestClient.java
```diff
@@ -0,0 +1,21 @@
...
+import io.vertx.core.impl.NoStackTraceTimeoutException;
+import jakarta.ws.rs.*;
+import jakarta.ws.rs.core.MediaType;
+import org.eclipse.microprofile.faulttolerance.Retry;
+
+import java.net.SocketTimeoutException;
+import java.time.temporal.ChronoUnit;
+
+@Retry(delay = 1, delayUnit = ChronoUnit.SECONDS, maxRetries = 5,
+        retryOn = {SocketTimeoutException.class, NoStackTraceTimeoutException.class})
```
kichasov: **Double retry layer: `@Retry` on the REST client multiplies with the dispatcher's outbox retries.**

The dispatcher already implements persistent retry/backoff (failed send → `PENDING` with `nextAttemptAt = now + retryDelay`, capped at `maxAttempts`). Stacking MicroProfile `@Retry(maxRetries=5)` on top means a single `notify()` can retry up to 5 times *synchronously* before the dispatcher even sees the failure — up to `5 × readTimeout` (15 min with the hardcoded 3-min timeout) of blocking, all inside the `@Transactional` dispatch loop holding row locks.

It also blurs ownership of retry semantics (which layer defines attempt count / backoff?).

For an outbox, prefer a single retry layer — the dispatcher. Drop `@Retry` from the REST client (or set `maxRetries = 0`) so each `dispatch()` pass fails fast and the dispatcher owns reschedule, backoff, attempt counting, and metrics.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/repositories/pg/jpa/OperatorEventRepository.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public long deleteByStatusAndCreatedAtBefore(OperatorEventStatus status, OffsetDateTime cutoff) {
+        return delete("status = ?1 and createdAt < ?2", status, cutoff);
+    }
+
+    /**
+     * Returns the most recent occurredAt timestamp for the given (classifier, type, role) tuple,
+     * used to populate previousRotatedAt in new outbox events.
+     */
+    public OffsetDateTime findPreviousOccurredAt(String classifierJson, String type, String role) {
```
kichasov: **`previousRotatedAt` is derived from the outbox table, which `cleanup()` purges by TTL — so it silently becomes `null` after `sentTtl`.**

`findPreviousOccurredAt()` reads `MAX(payload->>'occurredAt')` from `operator_event_outbox`, but `cleanup()` deletes `SENT` rows older than `sentTtl` (default `PT168H` = 7 days). If two rotations of the same `(classifier, type, role)` are more than `sentTtl` apart, the previous row is already gone and `previousRotatedAt` comes back `null` even though a prior rotation did happen. Effectively the field means "last rotation within the retention window", not "last rotation".

If the operator relies on `previousRotatedAt` for ordering / idempotency / gap detection, this is a correctness gap. Options: keep a durable "last rotation" timestamp outside the ephemeral outbox (e.g. alongside the DB record updated on password change), or explicitly document the field as best-effort within `sentTtl` so consumers don't depend on it. What does the operator use this field for?

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/OperatorEventDispatcher.java
kichasov, line 48: **No unit test for `OperatorEventDispatcher` — the failure-handling branches are untested.**

The dispatcher is the most branch-heavy part of this PR and owns delivery reliability: success → `SENT`+`sentAt`, 4xx → immediate `FAILED`, 5xx/timeout → `PENDING`+reschedule, `attempts > maxAttempts` → `FAILED`, plus attempt/error bookkeeping and metrics. Only `OperatorEventOutboxWriter` has a unit test; the dispatcher is exercised only by the happy-path IT.

Suggest an `OperatorEventDispatcherTest` (Mockito, mirroring `OperatorEventOutboxWriterTest`) with mocked `OperatorWebhook` / `OperatorEventRepository` / `OperatorEventMetrics`, covering: success→SENT, 4xx→FAILED (no retry), 5xx→PENDING+retry, attempts>maxAttempts→FAILED, and the metric increments. This would also pin down the attempt-counting semantics flagged separately.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/OperatorEventDispatcher.java
```diff
@@ -0,0 +1,123 @@
...
+            }
+            event.increaseAttempt();
+            event.setLastAttemptAt(OffsetDateTime.now());
+            operatorEventRepository.save(event);
+        }
+    }
+
+
+    private void scheduleRetryOrFail(OperatorEvent event, String errorMsg) {
+        int nextAttempt = event.getAttempts() + 1;
```
kichasov: **Off-by-one in attempt accounting: `maxAttempts = N` actually performs `N + 1` sends.**

`scheduleRetryOrFail()` computes `nextAttempt = event.getAttempts() + 1` *before* `event.increaseAttempt()` runs (the increment happens after the catch block). Walking it through with `maxAttempts = 10`: the 11th `notify()` is still executed and only marked `FAILED` after it fails — so a configured max of 10 yields 11 real HTTP sends.

Not severe, but the `attempts` / `maxAttempts` semantics are confusing. Suggest either incrementing `attempts` before the check, or using `>=`, and pinning the exact count with the dispatcher unit test.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/OperatorEventDispatcher.java
```diff
@@ -0,0 +1,123 @@
...
+                    scheduleRetryOrFail(event, msg);
+                }
+            } catch (Exception e) {
+                String msg = RestClientExceptionUtil.extractErrorMessage(e);
+                event.setLastError(msg);
+                scheduleRetryOrFail(event, msg);
+            }
+            event.increaseAttempt();
+            event.setLastAttemptAt(OffsetDateTime.now());
+            operatorEventRepository.save(event);
```
kichasov: **Redundant `merge` on already-managed entities.**

Events come from `claimPendingBatch()` via the same `EntityManager` within this `@Transactional` method, so they are managed — the `setStatus`/`setSentAt`/`increaseAttempt` mutations are flushed automatically at commit. The explicit `save()` → `merge()` per event is unnecessary work. Dropping it (relying on dirty-checking) is enough; keep only if you prefer the explicitness.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/DbBackupV2Service.java
```diff
@@ -1416,6 +1411,18 @@ protected void createLogicalDatabasesFromRestore(Restore restore,
                 newDatabase.setResources(newDatabase.getResources().stream().distinct().collect(Collectors.toList()));
                 encryption.encryptPassword(newDatabase);
                 databaseRegistryDbaasRepository.saveInternalDatabase(newDatabase.getDatabaseRegistry().getFirst());
+                for (DatabaseRegistry registry : newDatabase.getDatabaseRegistry()) {
+                    for (EnsuredUser eu : ensuredUsers) {
```
kichasov: **Possible duplicate `RESTORE_COMPLETED` events from the nested `registry × ensuredUsers` loop.**

An event is enqueued per `(DatabaseRegistry, EnsuredUser)` pair. If a database has multiple registries (multiple classifiers) and/or `ensuredUsers` repeats a `ROLE`, this emits several events for the same `(classifier, role)`. The operator should be idempotent, but this generates redundant outbox rows/webhooks. Contrast with `ROTATION_OCCURRED` in `DBaaService`, which enqueues once per `(registry, role)` without iterating users. Is exactly one event per `(classifier, role)` expected here? If so, dedupe by role before enqueueing.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/DBaaService.java
```diff
@@ -342,6 +345,11 @@ PasswordChangeResponse performChangePassword(List<DatabaseRegistry> databasesFor
                     response.putSuccessEntity(databaseRegistry.getClassifier(), new HashMap<>(ConnectionPropertiesUtils.getConnectionProperties(database.getConnectionProperties(), (String) cp.get(ROLE))));
                     encryption.encryptPassword(database, (String) cp.get(ROLE));
                     logicalDbDbaasRepository.getDatabaseRegistryDbaasRepository().saveInternalDatabase(databaseRegistry);
+                    operatorEventOutboxWriter.enqueue(
+                            OperatorEventType.ROTATION_OCCURRED,
```
kichasov: **`role` passed to `enqueue` may be `null`, silently disabling the previous-rotation lookup.**

`(String) cp.get(ROLE)` can be `null`. The payload handles that (`@JsonInclude(NON_NULL)` drops it), but `findPreviousOccurredAt(..., role=null)` runs `payload->>'userRole' = :role` with a null parameter, which is never true in SQL — so `previousRotatedAt` is always `null` for role-less events. Not a crash, just a silent miss. Worth making the contract explicit: either guarantee a non-null role before enqueueing, or document that null role skips the previous lookup.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/DBaaService.java
```diff
@@ -342,6 +345,11 @@ PasswordChangeResponse performChangePassword(List<DatabaseRegistry> databasesFor
...
+                            databaseRegistry.getClassifier(),
+                            databaseRegistry.getType(),
+                            (String) cp.get(ROLE));
```
kichasov: **[P1] Successful rotations can lose their outbox event (and their persisted new password) on a partial-batch failure.**

`changeUserPassword(...)` is `@Transactional`, and `performChangePassword()` runs in that same transaction. Inside the loop, each successful rotation performs an external adapter password change (`recreateUsers(...)`, line 335) and then persists the new encrypted password (`saveInternalDatabase`, line 347) and enqueues the outbox event (line 348). But after the loop, any failed database triggers `throw new PasswordChangeFailedException(...)` (line 245), which rolls back the **entire** transaction — including the `saveInternalDatabase` and outbox rows of the already-succeeded databases.

The external adapter-side password change has already happened and is **not** rolled back. So in a partially failed batch: (1) the aggregator DB reverts to the old credentials while the real password is new (credential desync), and (2) the operator is never notified, leaving `DatabaseSecret` consumers stale until a safety poll.

Guarantee commit for the successful subset — e.g. persist successful DB updates + their outbox events in a transaction that the final partial-failure response does not roll back (or make the operation truly all-or-nothing *before* any external password mutation).

dbaas/dbaas-aggregator/src/main/resources/application.properties
```diff
@@ -106,6 +106,18 @@ quarkus.smallrye-openapi.info-description=This documentation presents the REST A
...
+dbaas.operator.notification.enabled=${KUBERNETES_JWT_ENABLED:false}
+dbaas.operator.notification.callback-url=${DBAAS_OPERATOR_NOTIFICATION_CALLBACK_URL:http://dbaas-operator:8080}
+dbaas.operator.notification.jwt-audience=${DBAAS_OPERATOR_NOTIFICATION_JWT_AUDIENCE:dbaas-operator}
+dbaas.operator.notification.retry-delay=${DBAAS_OPERATOR_NOTIFICATION_RETRY_DELAY:PT10S}
+dbaas.operator.notification.max-attempts=${DBAAS_OPERATOR_NOTIFICATION_MAX_ATTEMPTS:10}
+dbaas.operator.notification.sent-ttl=${DBAAS_OPERATOR_NOTIFICATION_SENT_TTL:PT168H}
+dbaas.operator.notification.failed-ttl=${DBAAS_OPERATOR_NOTIFICATION_FAILED_TTL:PT720H}
+dbaas.operator.notification.connect-timeout=${DBAAS_OPERATOR_NOTIFICATION_CONNECT_TIMEOUT:PT30S}
+dbaas.operator.notification.read-timeout=${DBAAS_OPERATOR_NOTIFICATION_READ_TIMEOUT:PT30S}
+dbaas.operator.notification.invisibility-time=${DBAAS_OPERATOR_INVISIBILITY_TIME:PT1M}
```
kichasov: **[N1] `invisibility-time` (60s) has no margin over the max single-delivery time (`connect-timeout + read-timeout` = 60s) — a slow webhook can be re-dispatched as a duplicate.**

`claimPendingBatch` hides a claimed event for `invisibilityTime` by pushing `next_attempt_at = NOW() + invisibilityTime`. With the defaults `invisibility-time=PT1M` and `connect-timeout`/`read-timeout`=`PT30S` each, the worst-case `notify()` (connect + read) is also ~60s — equal to the lease. If a delivery runs close to its timeout, the invisibility window expires at about the same moment the HTTP call is still in flight, so the next `dispatch()` tick (or another replica) re-claims the same event and sends it again. The operator is keyed by `eventId` so it can dedupe, but this is avoidable duplicate traffic/load.

The safe invariant is `invisibilityTime > connectTimeout + readTimeout + buffer`. Suggest either bumping the default (e.g. `PT2M`), deriving it from the timeouts, or at minimum documenting the relationship so operators don't set it too low.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/OperatorEventStatusUpdater.java
```diff
@@ -0,0 +1,97 @@
...
+     */
+    @Transactional(REQUIRES_NEW)
+    public void markRetry(UUID id, String msg) {
+        OperatorEvent event = operatorEventRepository.findById(id);
+        if (event == null) {
+            log.warn("markRetry: event {} not found, skipping", id);
+            return;
+        }
+        event.setLastError(msg);
+        if (event.getAttempts() >= rotationProperty.maxAttempts()) {
```
kichasov, line 76: **[L1] `maxAttempts = N` results in ~N+1 actual `notify()` calls.**

`deliver()` calls `notify()` *before* `markRetry()`, and `markRetry` checks `getAttempts() >= maxAttempts` (where `getAttempts()` is the count before this attempt's `stamp()`). So with `maxAttempts=10`, the event is still delivered on the pass where `attempts==10` and only then marked `FAILED` — ~11 real HTTP sends. Minor, but the `attempts`/`maxAttempts` contract is easy to misread; worth pinning the exact count with a unit test on `OperatorEventStatusUpdater` (attempts-exhaustion → `FAILED`, no further retry).

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/DBaaService.java
```diff
@@ -341,7 +345,9 @@ PasswordChangeResponse performChangePassword(List<DatabaseRegistry> databasesFor
 
                     response.putSuccessEntity(databaseRegistry.getClassifier(), new HashMap<>(ConnectionPropertiesUtils.getConnectionProperties(database.getConnectionProperties(), (String) cp.get(ROLE))));
                     encryption.encryptPassword(database, (String) cp.get(ROLE));
-                    logicalDbDbaasRepository.getDatabaseRegistryDbaasRepository().saveInternalDatabase(databaseRegistry);
+                    commitPasswordRotation(databaseRegistry, databaseRegistry.getClassifier(),
```
kichasov: **`commitPasswordRotation` is invoked once per role inside the `for (cp : connectionProperties)` loop, so a multi-role database enqueues N identical `(classifier, type)` ROTATION_OCCURRED events.** Two issues after dropping `role` from the event:

1. **The "avoid redundant requests" goal isn't met.** The outbox still gets N rows and `OperatorEventDispatcher` sends N identical webhooks per rotation — removing `role` only made the duplicates indistinguishable, it didn't reduce their count.
2. **`previousRotatedAt` regression.** `findPreviousOccurredAt(classifier, type)` (no longer filtered by role) returns `MAX(occurredAt)` over *all* events for that `(classifier, type)`. Since each `commitPasswordRotation` runs in its own `QuarkusTransaction.requiringNew()` (committed immediately) and the roles are processed sequentially, the lookup for role #2..N matches the sibling event role #1 just committed → their `previousRotatedAt` ends up ≈ `now` (this same rotation) instead of the actual prior rotation. Only the first event is correct. There is no unique constraint on `operator_event_outbox`, so nothing fails — the bug is silent.

Hoist `commitPasswordRotation` out of the per-role loop and enqueue **once per database** after all roles have rotated. That fixes both the duplicate webhooks and the `previousRotatedAt` contamination.

PR #524, opened 2026-06-23
feat: metrics for rule, polling, and namespacebinding CRs

TheHollowMonarch:
Summary:
- Adds current-state CR metrics for phase, conditions, observed generation lag, deletion/finalizer state, NamespaceBinding state, SecretClaim rotation/not-found state, and balancing-rule desired vs applied target counts.
- Registers the custom resource metrics collector with the manager metrics registry.
- Extends existing operation metrics to cover DatabaseSecretClaim aggregator calls and reconcile trigger sources.
- Splits balancing-rule reconcile trigger metrics by rule kind.
- Updates the Grafana dashboard with CR health, SecretClaim, balancing-rule, NamespaceBinding, and deletion-state panels.

### DIFF
--- a/dbaas-operator/cmd/main.go
+++ b/dbaas-operator/cmd/main.go
@@ -221,6 +221,7 @@ func main() {
 
 	// ── Ownership resolver ────────────────────────────────────────────────────
 	ownershipResolver := ownership.NewOwnershipResolver(cloudNamespace, mgr.GetClient())
+	controller.RegisterResourceMetrics(mgr.GetClient(), ownershipResolver, cloudNamespace)
 	if err := mgr.Add(&ownershipWarmupRunnable{resolver: ownershipResolver}); err != nil {
 		setupLog.Errorf("Failed to register ownership warmup runnable: %v", err)
 		os.Exit(1)
--- a/dbaas-operator/helm-templates/dbaas-operator/dashboards/dbaas-operator-dashboard.json
+++ b/dbaas-operator/helm-templates/dbaas-operator/dashboards/dbaas-operator-dashboard.json
@@ -355,7 +355,7 @@
                    {
                        "id":  8,
                        "type":  "row",
-                       "title":  "Dashboard-Only Resource Health Substitutes",
+                       "title":  "Historical Failure Activity",
                        "collapsed":  false,
                        "gridPos":  {
                                        "h":  1,
@@ -368,7 +368,7 @@
                        "id":  9,
                        "type":  "timeseries",
                        "title":  "Recent Failed Operations by Owner",
-                       "description":  "Dashboard-only substitute for resource phase health. Shows recent failed operations by owner bucket instead of exact current CR phase counts.",
+                       "description":  "Historical failed operations by owner bucket. Use the CR Health Overview row for exact current CR state.",
                        "gridPos":  {
                                        "h":  8,
                                        "w":  12,
@@ -415,7 +415,7 @@
                        "id":  10,
                        "type":  "timeseries",
                        "title":  "Sustained Failure Signals",
-                       "description":  "Dashboard-only substitute for stuck resources. Shows failure signals that continue over the recent window; it is event-based, not an exact stuck CR count.",
+                       "description":  "Failure signals that continue over the recent window. This is event-based history; use CR phase and condition panels for exact current stuck CRs.",
                        "gridPos":  {
                                        "h":  8,
                                        "w":  12,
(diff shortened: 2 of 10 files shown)


kichasov:
**[P1] Current-state gauges double-count with multiple operator replicas**

The resource-metrics collector is registered unconditionally on every pod (`RegisterResourceMetrics` in `cmd/main.go` — not gated on leader election), so every replica emits identical current-state series for the same CRs, differing only by the `pod`/`instance` target labels. The PodMonitor scrapes all pods and the `namespace="#namespace#"` filter doesn't separate them (replicas share the namespace), so any panel that does `sum(...)` over these gauges multiplies every count by the replica count.

Affected panels: **20, 21, 22, 28, 29, 30** (CR phase counts, balancing desired/applied targets, etc.).

Recommended fix — dedupe per-pod **before** aggregating, in the dashboard queries:

```promql
sum by (kind, phase) (
  max without (pod, instance) (dbaas_resource_phase{namespace="#namespace#"})
)
```

Replica values are identical, so `max without (pod, instance)` collapses them to a single series per CR. Apply the same `max without (pod, instance)` wrapper to the other affected current-state gauges.

Keep the **operational counters per-pod** (`dbaas_aggregator_requests_total`, reconcile-trigger counters, `dbaas_resource_collector_success`) — those should reflect each replica's health, so they must not be deduped/leader-gated. A leader-only export of the current-state gauges is a valid alternative, but it introduces a metrics gap on failover and makes non-leader pods look empty, so query-side dedup is preferred.

kichasov:
**[P2] Expected `DatabaseNotFound` (CORE-DBAAS-4006) is recorded as `server_error`**

In `DatabaseSecretClaimReconciler.Reconcile`, `recordAggregatorCall(controllerDS, operationGetDatabase, aggStart, err)` is called with the raw error **before** `handleAggregatorErr`. `aggregatorResult` only special-cases auth and spec-rejection, so a `DatabaseNotFound` error falls through to `resultServerError`:

```go
if errors.As(err, &aggErr) {
    if aggErr.IsAuthError()     { return resultAuthError }
    if aggErr.IsSpecRejection() { return resultSpecRejection }
    return resultServerError   // <- DatabaseNotFound lands here
}
```

But `handleAggregatorErr` itself treats `404 + CORE-DBAAS-4006` as a normal transient wait ("BackingOff / DatabaseNotFound, DB not yet registered" — tracked via `FirstNotFoundAt`, escalated only after `databaseNotFoundTimeout`). So while a database is still provisioning, every poll increments `dbaas_aggregator_requests_total{operation="get_database_by_classifier", result="server_error"}`, surfacing normal waiting as a server failure in "Recent Failed Operations" / "Sustained Failure Signals" (and any alert on the server_error rate).

Fix — classify it distinctly, reusing the existing predicate:

```go
// in aggregatorResult, before the server_error fallthrough:
if aggErr.IsDatabaseNotFound() {
    return resultDatabaseNotFound
}
// ...
const resultDatabaseNotFound = "database_not_found" // or "waiting"
```

and exclude `database_not_found` from the failure panels/alerts. Consider this **P1** if `server_error` feeds paging alerts — it would page on every normal provisioning wait.

kichasov:
**[Medium] Duplicate status conditions can fail the whole metrics scrape**

`collectOperatorStatus` emits one `dbaas_resource_condition` series per `status.Conditions[i]` with labels `(kind, resource_namespace, name, condition, status, reason)`. If a single CR ever carries two conditions with the same `(type, status, reason)`, the collector sends two metrics with an identical descriptor + label set within one `Collect`. The registry detects this during `Gather` as a duplicate and errors out — dropping this collector's metrics for that scrape (potentially the whole `/metrics` response), not just the one offending series. kube-state-metrics guards against exactly this.

Conditions are normally unique by `Type`, but the metrics endpoint shouldn't be one operator-status bug away from going dark. Cheap defensive fix — dedupe before emitting (keep one per `Type`):

```go
seen := make(map[string]struct{}, len(status.Conditions))
for _, condition := range status.Conditions {
    if _, dup := seen[condition.Type]; dup {
        continue
    }
    seen[condition.Type] = struct{}{}
    ch <- prometheus.MustNewConstMetric(resourceConditionDesc, /* ... */)
}
```

A `Collect`-level test that feeds a CR with duplicate conditions (e.g. `testutil.CollectAndCount`) would lock this in.

kichasov:
**Minor findings (Low / Nit)**

Non-blocking follow-ups from reviewing the resource-metrics collector; grouping them here to keep the thread tidy.

**[Low] `reason` label cardinality on `dbaas_resource_condition`.** The metric carries `reason` as a label. That's fine as long as `condition.Reason` is a bounded, CamelCase enum (the k8s convention). If anything ever puts dynamic/free-form text into `Reason`, the time-series cardinality explodes. Please confirm `Reason` is bounded for every CR that exposes `OperatorStatus`. (Good call leaving `message` out of the labels.)

**[Low] `collectPermanentBalancingRules` with an empty operator namespace.** It emits `dbaas_resource_collector_success{kind="PermanentBalancingRule"}=1` and *then* does `if c.operatorNamespace == "" { return }`, so it reports success while emitting no series; and `client.List(ctx, list, client.InNamespace(""))` lists across all namespaces in that case. Prefer checking the empty namespace **before** the `List`/success emission, so the success signal matches what was actually collected and the list stays scoped.

**[Low] Reconcile trigger is computed (and consumed) before the ownership check.** In `DatabaseSecretClaimReconciler.Reconcile`, `trigger := r.triggerForSecretClaim(key, s)` runs before `checkOwnership`, and `consumeRotationTrigger`/`consumeSiblingTrigger`/`consumeBindingTrigger` mutate the dedup maps. For a CR this instance doesn't own, the trigger is consumed but never recorded (`recordReconcileTrigger` only runs for owned CRs), so an ownership flap can swallow a trigger. Move the trigger computation after the `if !owned { ... }` early-return. (Minor related leak: `rotationTriggerValues` is only cleared on delete, not in the `!owned` branch.)

**[Nit] `dbaas_secret_claim_not_found_age_seconds` is computed at scrape time.** Exposing a derived age means the value drifts between scrapes. It's more idiomatic (and matches what you already do for rotation) to export the raw timestamp — e.g. `dbaas_secret_claim_first_not_found_timestamp_seconds` — and compute the age in PromQL via `time() - <metric>`.

**[Nit] Tests only cover the pure helpers.** `generationLag`, `namespaceBindingState`, and the target-count helpers are covered, but there's no `Collect`-level test for the collector itself: the `ownsNamespace` filtering, `collector_success=0` on a `List` error, and the duplicate-condition dedup (see the separate Medium comment). A `prometheus/testutil.CollectAndCount` / `CollectAndCompare` test against a fake client would cover these.

Issue #143, opened 2025-10-01
Documentation: Add Installation guide


RomanB89:
Please add missing installation documentation for DBaaS

PR #551, opened 2026-07-03
feat(logging): structured NDJSON logging migration

al-matushkin:
## Summary

Migrates **qubership-dbaas** to structured NDJSON logging (platform logging guide / QIP #143), covering both runtime components:

- **dbaas-aggregator** (Quarkus/SLF4J): `quarkus-logging-json`, `StructuredLog` MDC helper, `LOG_FORMAT` / `QUARKUS_PROFILE` Helm wiring
- **dbaas-operator** (Go): `internal/logformat` + `internal/logfields`, `LOG_FORMAT` Helm default `json`

Generated with the `qubership-ndjson-logging-migration` skill. Full coverage ledger: `.ndjson-migration-report.md`.

## Completion gate

| Check | Before | After |
|-------|--------|-------|
| Java `{}` SLF4J calls in `src/main/java` | 632 | 0 |
| Go production formatted log args (`%v`/`%d`/`%q`) | 62 | 0 |
| Logged preformatted `log.warn(message)` / `log.error(msg)` | 10 | 0 |

## Validation

| Check | Result |
|-------|--------|
| `GOWORK=off go build ./cmd/` | PASS |
| `GOWORK=off go test` (api, cmd, client, ownership, poller) | PASS |
| `go test ./internal/controller/...` | FAIL — envtest needs kubebuilder etcd (environmental) |
| Maven `compile` (aggregator) | Not verified locally — GitHub Packages 401 |
| NDJSON startup smoke (`LOG_FORMAT=json`) | PASS |

## Follow-up (optional)

- **Exception mappers (12 sites):** `log.warn/error(WARNING_MESSAGE, …)` in `controller/error/*` and `Utils.java` still use the shared `{}` template constant; `{}` completion gate is 0 because placeholders live in the constant. Consider centralizing via `StructuredLog` in a follow-up PR.

## Test plan

- [ ] CI: aggregator Maven build with GitHub Packages credentials
- [ ] CI: operator unit + integration tests
- [ ] Deploy with `LOG_FORMAT=json` (default) and verify NDJSON in collector
- [ ] Rollback path: set `LOG_FORMAT=text` / `QUARKUS_PROFILE=text` for legacy format

Made with [Cursor](https://cursor.com)

### DIFF
--- a/dbaas-operator/internal/logformat/logformat.go
+++ b/dbaas-operator/internal/logformat/logformat.go
@@ -1,0 +1,78 @@
+package logformat
+
+import (
+	"bytes"
+	"encoding/json"
+	"os"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/netcracker/qubership-core-lib-go/v3/logging"
+)
+
+var kvSuffix = regexp.MustCompile(`(?:^|\s)([A-Za-z_][A-Za-z0-9_.-]*)=((?:'[^']*')|(?:"[^"]*")|(?:\S+))`)
+
+// Init selects NDJSON or legacy text output from LOG_FORMAT (default json).
+func Init() {
+	Configure()
+}
+
+// Configure selects NDJSON or legacy text output from LOG_FORMAT (default json).
+func Configure() {
+	switch strings.ToLower(strings.TrimSpace(os.Getenv("LOG_FORMAT"))) {
+	case "", "json":
+		logging.SetLogFormat(ndjsonFormat)
+	case "text":
+		// Keep library default bracket text formatter.
+	default:
+		logging.SetLogFormat(ndjsonFormat)
+	}
+}
+
+func ndjsonFormat(r *logging.Record) []byte {
+	message, fields := splitMessageFields(r.Message)
+	record := map[string]any{
+		"time":    r.Time.UTC().Format(time.RFC3339Nano),
+		"level":   strings.ToUpper(r.Lvl.String()),
+		"message": message,
+	}
+	if r.PackageName != "" {
+		record["class"] = r.PackageName
+	}
+	putContextField(record, "request_id", logging.GetValueOrPlaceholder(r.Ctx, logging.RequestIdContextName))
+	putContextField(record, "tenant_id", logging.GetValueOrPlaceholder(r.Ctx, logging.TenantContextName))
+	putContextField(record, "x_channel_request_id", logging.GetValueOrPlaceholder(r.Ctx, logging.ChannelRequestIdContextName))
+	for k, v := range fields {
+		record[k] = v
+	}
+	var buf bytes.Buffer
+	enc := json.NewEncoder(&buf)
+	enc.SetEscapeHTML(false)
+	_ = enc.Encode(record)
+	return buf.Bytes()
+}
+
+func putContextField(record map[string]any, key, value string) {
+	if value != "" && value != "-" {
+		record[key] = value
+	}
+}
+
+func splitMessageFields(message string) (string, map[string]string) {
+	fields := make(map[string]string)
+	matches := kvSuffix.FindAllStringSubmatchIndex(message, -1)
+	if len(matches) == 0 {
+		return strings.TrimSpace(message), fields
+	}
+	cut := len(message)
+	for _, m := range matches {
+		if m[0] < cut {
+			cut = m[0]
+		}
+		name := message[m[2]:m[3]]
+		raw := message[m[4]:m[5]]
+		fields[name] = strings.Trim(raw, `"'`)
+	}
+	return strings.TrimSpace(message[:cut]), fields
+}
--- a/dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/controller/error/ForbiddenExceptionMapper.java
+++ b/dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/controller/error/ForbiddenExceptionMapper.java
@@ -1,4 +1,5 @@
 package com.netcracker.cloud.dbaas.controller.error;
+import com.netcracker.cloud.dbaas.logging.StructuredLog;
 
 import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.Response;
--- a/dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/controller/error/MultiValidationExceptionMapper.java
+++ b/dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/controller/error/MultiValidationExceptionMapper.java
@@ -1,4 +1,5 @@
 package com.netcracker.cloud.dbaas.controller.error;
+import com.netcracker.cloud.dbaas.logging.StructuredLog;
 
 import com.netcracker.cloud.core.error.runtime.ErrorCodeException;
 import com.netcracker.cloud.dbaas.exceptions.MultiValidationException;
@@ -24,14 +25,9 @@ public class MultiValidationExceptionMapper implements ExceptionMapper<MultiVali
 
     @Override
     public Response toResponse(MultiValidationException e) {
-        log.warn("{} happened during request to {}. Validation errors: {}", e.getClass().getSimpleName(), uriInfo.getPath(),
-                e.getValidationExceptions().stream().map(ErrorCodeException::getMessage).collect(Collectors.joining("\n")));
+        StructuredLog.warn(log, "happened during request to . Validation errors:", "arg0", e.getClass().getSimpleName(), "arg1", uriInfo.getPath(), "status", e.getValidationExceptions().stream().map(ErrorCodeException::getMessage).collect(Collectors.joining("\n")));
         Response.Status status = Response.Status.BAD_REQUEST;
-        return buildResponse(status,
-                () -> tmfResponseBuilder(e, status)
-                        .errors(e.getValidationExceptions().stream().map(ve ->
-                                tmfErrorBuilder(ve, status).build()).collect(Collectors.toList()))
-                        .build());
+        return buildResponse(status);
 
     }
 }
--- a/dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/controller/v3/AggregatedDatabaseAdministrationNoNamespaceControllerV3.java
+++ b/dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/controller/v3/AggregatedDatabaseAdministrationNoNamespaceControllerV3.java
@@ -1,4 +1,5 @@
 package com.netcracker.cloud.dbaas.controller.v3;
+import com.netcracker.cloud.dbaas.logging.StructuredLog;
 
 import com.netcracker.cloud.core.error.rest.tmf.TmfErrorResponse;
 import com.netcracker.cloud.dbaas.controller.abstact.AbstractController;
@@ -70,7 +71,7 @@ public class AggregatedDatabaseAdministrationNoNamespaceControllerV3 extends Abs
     public Response getDatabasesByName(@PathParam("dbname") String dbname,
                                        @QueryParam(NAMESPACE_PARAMETER) String namespace,
                                        @QueryParam("withDecryptedPassword") @DefaultValue("false") Boolean withDecryptedPassword) {
-        log.info("Get databases with name {}. Query params: namespace {}, withDecryptedPassword {}", dbname, namespace, withDecryptedPassword);
+StructuredLog.info(log, "Get databases with name. Query params: namespace, withDecryptedPassword", "dbname", dbname, "namespace", namespace, "withDecryptedPassword", withDecryptedPassword);
         List<DatabaseRegistry> bdRegistries = databaseRegistryDbaasRepository.findAnyLogDbTypeByNameAndOptionalParams(dbname, namespace);
         if (withDecryptedPassword) {
             bdRegistries.forEach(dbRegistry -> dBaaService.decryptPassword(dbRegistry.getDatabase()));
@@ -126,7 +127,7 @@ public class AggregatedDatabaseAdministrationNoNamespaceControllerV3 extends Abs
             throw new BadRequestException("Query parameter 'namespaces' must be not empty list of strings");
         }
 
-        log.info("Received request to drop all logical databases in {} namespaces {}", namespaces.size(), namespaces);
+StructuredLog.info(log, "Received request to drop all logical databases in namespaces", "count", namespaces.size(), "namespaces", namespaces);
 
         assertNotProdMode();
 
@@ -134,7 +135,7 @@ public class AggregatedDatabaseAdministrationNoNamespaceControllerV3 extends Abs
 
         if (logicalDatabasesAmount == 0) {
 
-            log.info("Namespaces {} are empty, dropping is not needed", namespaces);
+StructuredLog.info(log, "Namespaces are empty, dropping is not needed", "namespaces", namespaces);
 
             return Response.ok(String.format("All %s namespaces %s do not contain any logical databases", namespaces.size(), namespaces)).build();
         }
@@ -158,7 +159,7 @@ public class AggregatedDatabaseAdministrationNoNamespaceControllerV3 extends Abs
     @RolesAllowed(NAMESPACE_CLEANER)
     public Response getMarkedForDropDatabases(@Parameter(description = "List of namespaces by which marked for drop databases is needed to return", required = true)
                                               List<String> namespaces) {
-        log.info("Receive request to get marked for drop databases in '{}' namespaces", namespaces);
+StructuredLog.info(log, "Receive request to get marked for drop databases in '' namespaces", "namespaces", namespaces);
         if (CollectionUtils.isEmpty(namespaces)) {
             throw new RequestValidationException(ErrorCodes.CORE_DBAAS_4043,
                     ErrorCodes.CORE_DBAAS_4043.getDetail("Should be at least one namespace specified"),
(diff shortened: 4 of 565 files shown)


dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/controller/v3/AggregatedDatabaseAdministrationNoNamespaceControllerV3.java
```diff
@@ -211,47 +211,6 @@ public Response cleanupMarkedForDropDatabases(CleanupMarkedForDropRequest reques
...
-            id = (sinceId == null || sinceId.isBlank()) ? new UUID(0L, 0L) : UUID.fromString(sinceId);
-        } catch (IllegalArgumentException e) {
-            throw new BadRequestException("Query parameter 'sinceId' must be a UUID");
-        }
-        List<ChangedDatabaseResponse> items = databaseRegistryDbaasRepository.findChangedSince(ts, id, limit).stream()
-                .map(ChangedDatabaseResponse::new)
-                .toList();
-        log.debug("Returned {} changed databases since ({}, {})", items.size(), ts, id);
-        return Response.ok(new ChangedDatabasesResponse(items, highWaterMark)).build();
+            @APIResponse(responseCode = "200", description = "Changed databases and the current high-water mark", content = @Content(schema = @Schema(implementation = ChangedDatabasesResponse.class))));
```
kichasov, line 214: **P0 — compile break + removed endpoint.** The `GET /changed` handler body was deleted here; only a dangling `@Operation` and a malformed `@APIResponses` remain (`@APIResponse(...)))` has unbalanced parentheses and annotates nothing), so `dbaas-aggregator` no longer compiles. Beyond the syntax break, this drops the `GET /api/v3/dbaas/databases/changed` endpoint entirely — which the dbaas-operator rotation poller depends on (`internal/client/aggregator_client.go:327`, `GetChangedSince` → `ChangedDatabasesResponse`). Restore the complete method (the `@GET`/path annotation, signature, and body), not just the doc annotations.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/DBBackupsService.java
```diff
@@ -246,33 +237,21 @@ protected void deleteAllNamespaceBackupsInNamespacesByPortions(Set<String> names
...
                 allFailedDeletedNamespaceBackupIdsAmount += failedDeleteNamespaceBackupsIds.size();
 
-                log.info("""
-                                Finished deletion of {} namespace backups by portion with number {}, successfully deleted {} namespace backups with ids {}, \
-                                failed deleted {} ones with ids {}""",
-
-                        namespaceBackupsIds.size(), portionNumber, successfullyDeletedNamespaceBackupsIds.size(),
-                        successfullyDeletedNamespaceBackupsIds, failedDeleteNamespaceBackupsIds.size(), failedDeleteNamespaceBackupsIds
-                );
+                StructuredLog.info(log, """ Finished deletion of namespace backups by portion with number , successfully deleted namespace backups with ids , \ failed deleted ones with ids """, "namespace", namespaceBackupsIds.size(), "portionNumber", portionNumber, "namespace", successfullyDeletedNamespaceBackupsIds.size(), "successfullyDeletedNamespaceBackupsIds", successfullyDeletedNamespaceBackupsIds, "namespace", failedDeleteNamespaceBackupsIds.size(), "failedDeleteNamespaceBackupsIds", failedDeleteNamespaceBackupsIds);
```
kichasov, line 240: **P0 — compile break (illegal text block).** The conversion collapsed a multi-line message into a single-line Java text block: `StructuredLog.info(log, """ Finished deletion ... """, ...)`. A text block's opening `"""` must be immediately followed by a line terminator, so this is `illegal text block open delimiter sequence` and `mvn compile` fails. Use a normal `"..."` string literal. (This call also carries a stray `\`, `argN`-style keys, and a duplicated `namespace` key.) Same defect at `DeletionService.java:359`.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/DeletionService.java
```diff
@@ -359,30 +356,18 @@ private void cleanupAllLogicalDatabasesInNamespacesByPortions(Set<String> namesp
...
                 allFailedDeletedLogicalDatabaseIdsAmount += failedDeleteLogicalDatabasesIds.size();
 
-                log.info("""
-                                Finished deletion of {} logical databases by portion with number {}, successfully deleted {} logical databases with ids {}, \
-                                failed deleted {} ones with ids {}""",
-
-                        logicalDatabasesIds.size(), portionNumber, successfullyDeletedLogicalDatabasesIds.size(),
-                        successfullyDeletedLogicalDatabasesIds, failedDeleteLogicalDatabasesIds.size(), failedDeleteLogicalDatabasesIds
-                );
+                StructuredLog.info(log, """ Finished deletion of logical databases by portion with number , successfully deleted logical databases with ids , \ failed deleted ones with ids """, "arg0", logicalDatabasesIds.size(), "portionNumber", portionNumber, "arg2", successfullyDeletedLogicalDatabasesIds.size(), "successfullyDeletedLogicalDatabasesIds", successfullyDeletedLogicalDatabasesIds, "arg4", failedDeleteLogicalDatabasesIds.size(), "failedDeleteLogicalDatabasesIds", failedDeleteLogicalDatabasesIds);
```
kichasov, line 359: **P0 — compile break (illegal text block).** Same as `DBBackupsService.java:240`: the message was collapsed into a single-line text block (`""" ... """`), which is not valid Java (`illegal text block open delimiter sequence`) — `mvn compile` fails. Convert it to a normal string literal. (Also carries `arg0`/`arg2`/`arg4` placeholder keys.)

dbaas/dbaas-aggregator/src/main/java/db/migration/postgresql/V1_008__TransformClassifier.java
```diff
@@ -7,7 +7,7 @@
 import lombok.Getter;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
-import lombok.extern.slf4j.Slf4j;
+import com.netcracker.cloud.dbaas.logging.StructuredLog;
```
kichasov, line 10: **P0 — compile break (missing `@Slf4j` import).** This file keeps `@Slf4j` on the class (line 23) and still calls `log.info(...)`, but the conversion removed `import lombok.extern.slf4j.Slf4j;` — the import block here only added the `StructuredLog` import. Without the Lombok import, `@Slf4j` won't resolve (`cannot find symbol: class Slf4j`), so once the earlier syntax errors are fixed this file also stops the build. Restore `import lombok.extern.slf4j.Slf4j;`.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/controller/error/MultiValidationExceptionMapper.java
```diff
@@ -24,14 +25,9 @@ public class MultiValidationExceptionMapper implements ExceptionMapper<MultiVali
...
-        log.warn("{} happened during request to {}. Validation errors: {}", e.getClass().getSimpleName(), uriInfo.getPath(),
-                e.getValidationExceptions().stream().map(ErrorCodeException::getMessage).collect(Collectors.joining("\n")));
+        StructuredLog.warn(log, "happened during request to . Validation errors:", "arg0", e.getClass().getSimpleName(), "arg1", uriInfo.getPath(), "status", e.getValidationExceptions().stream().map(ErrorCodeException::getMessage).collect(Collectors.joining("\n")));
         Response.Status status = Response.Status.BAD_REQUEST;
-        return buildResponse(status,
-                () -> tmfResponseBuilder(e, status)
-                        .errors(e.getValidationExceptions().stream().map(ve ->
-                                tmfErrorBuilder(ve, status).build()).collect(Collectors.toList()))
-                        .build());
+        return buildResponse(status);
```
kichasov, line 30: **P0 — compile break + lost data.** `buildResponse(status)` has no matching overload: `controller/error/Utils.java` declares only two-arg forms (`buildResponse(StatusType, Supplier)` and `buildResponse(Status, Supplier)`), so this won't compile. The conversion also dropped the supplier `() -> tmfResponseBuilder(e, status).errors(...).build()`, so even after it compiles the TMF response loses the entire validation-errors list. Restore the two-arg call with the errors builder. (Separately, the `StructuredLog.warn` on line 28 uses `arg0`/`arg1` keys and files the joined validation messages under a `status` key.)

kichasov (commented):
**P0 review — `dbaas-aggregator` does not compile as-is, so this PR cannot be merged.** The auto-conversion introduced four blocking issues, flagged inline below (in addition to the earlier structured-logging review notes about dropped throwables, `argN` field keys, and duplicate keys).

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/DbBackupV2Service.java
```diff
@@ -245,8 +246,7 @@ protected void startBackup(Backup backup) {
                                         logicalBackup.setStatus(BackupTaskStatus.RETRYABLE_FAIL);
                                     }
                                     logicalBackup.setErrorMessage(extractErrorMessage(throwable));
-                                    log.error("Logical backup failed: logicalBackup={}, error={}",
-                                            logicalBackup.getId(), logicalBackup.getErrorMessage(), throwable);
+                                    StructuredLog.error(log, "Logical backup failed: logicalBackup=, error=", "backup", logicalBackup.getId(), "backup", logicalBackup.getErrorMessage());
```
kichasov, line 249: **Dropped throwable → lost stack trace.** The original `log.error("Logical backup failed: ...", id, msg, throwable)` passed the exception as the final SLF4J arg; this conversion uses the no-throwable `StructuredLog.error` overload, so the stack trace is gone. `StructuredLog` has an `error(logger, message, Throwable, fields...)` overload — use it. Across the aggregator diff, 23 removed `error`/`warn` calls passed a throwable but only ~44 conversions use the throwable overload, so this needs a sweep. (This call also has two `"backup"` keys — see the duplicate-key note.)

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/DbBackupV2Service.java
```diff
@@ -569,8 +576,7 @@ public void deleteBackup(String backupName, boolean force) {
         if (status != BackupStatus.COMPLETED &&
                 status != BackupStatus.FAILED &&
                 status != BackupStatus.DELETED) {
-            log.error("Backup has invalid status={}. Only {}, {} or {} backups can be processed.",
-                    status, BackupStatus.COMPLETED, BackupStatus.FAILED, BackupStatus.DELETED);
+            StructuredLog.error(log, "Backup has invalid status=. Only , or backups can be processed", "status", status, "backup", BackupStatus.COMPLETED, "backup", BackupStatus.FAILED, "backup", BackupStatus.DELETED);
```
kichasov, line 579: **Duplicate MDC keys overwrite each other.** `"backup", COMPLETED, "backup", FAILED, "backup", DELETED` — `StructuredLog.putFields` does `MDC.put(key, ...)` per pair, so the later `backup` values overwrite the earlier ones and only `DELETED` survives; `COMPLETED`/`FAILED` and the `status` value are lost. 23 `StructuredLog` calls in the aggregator carry a repeated key like this. Give each value a distinct, meaningful key.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/DBBackupsService.java
```diff
@@ -81,9 +82,9 @@ private Predicate<DatabaseRegistry> notMarkedForDrop() {
 
     @Transactional
     public NamespaceBackupDeletion deleteBackup(NamespaceBackup backupToDelete) throws NamespaceBackupDeletionFailedException, IllegalArgumentException {
-        log.info("Start delete backup for id {} in namespace {}", backupToDelete.getId(), backupToDelete.getNamespace());
+StructuredLog.info(log, "Start delete backup for id in namespace", "arg0", backupToDelete.getId(), "namespace", backupToDelete.getNamespace());
```
kichasov, line 85: **Placeholder field keys defeat structured logging.** `"arg0", backupToDelete.getId()` — the generator fell back to `arg0/arg1/...`. There are ~159 of these across the aggregator diff. You cannot filter or aggregate a log store on `arg0`, which is the whole point of QIP #143; they need meaningful snake_case names (`backup_id`, `namespace`, ...). The message text is also left mangled by the placeholder stripping (dangling `=`, empty `, ,` gaps) — worth a cleanup pass over the generated messages.

dbaas-operator/internal/logformat/logformat.go
```diff
@@ -0,0 +1,78 @@
...
+		"message": message,
+	}
+	if r.PackageName != "" {
+		record["class"] = r.PackageName
+	}
+	putContextField(record, "request_id", logging.GetValueOrPlaceholder(r.Ctx, logging.RequestIdContextName))
+	putContextField(record, "tenant_id", logging.GetValueOrPlaceholder(r.Ctx, logging.TenantContextName))
+	putContextField(record, "x_channel_request_id", logging.GetValueOrPlaceholder(r.Ctx, logging.ChannelRequestIdContextName))
+	for k, v := range fields {
+		record[k] = v
```
kichasov, line 47: **Format→regex round-trip can corrupt fields (Go).** Fields are serialized as ` key=value` onto the message string (`logfields.Format`) and re-parsed here by regex. (1) Unquoted values with whitespace get truncated — `logfields.Format("done", "path", "/a b")` yields `path=/a` plus a spurious `b` field; and any ordinary message that happens to contain `k=v` (a URL `?a=b`, `where id=5`) is partly promoted to fields and stripped from `message`. (2) `record[k] = v` merges parsed fields *after* the reserved keys with no guard, so a field named `time`/`level`/`message`/`class`/`request_id` silently overwrites the real one. Prefer emitting typed fields structurally, or at least quote values and whitelist/namespace the merged keys. (The current operator call sites are safe, but the helper is a foot-gun for future callers.)

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/logging/StructuredLog.java
```diff
@@ -0,0 +1,89 @@
...
+    public static void error(Logger logger, String message, Throwable throwable, Object... fields) {
+        withFields(logger, message, fields, (l, m) -> l.error(m, throwable));
+    }
+
+    @FunctionalInterface
+    private interface LogAction {
+        void log(Logger logger, String message);
+    }
+
+    private static void withFields(Logger logger, String message, Object[] fields, LogAction action) {
```
kichasov, line 51: Three notes on the helper. (1) **Level not checked** — `withFields` runs `putFields` + `stringify` on every argument and does MDC put/remove before the log call, so `StructuredLog.debug(...)` on a disabled level still pays that cost; the original `{}` calls were lazy and level-guarded. Consider an `isDebugEnabled`/`isTraceEnabled` guard for the hot levels. (2) **Field visibility** — arbitrary snake_case keys go into MDC, but Quarkus only promotes *declared* MDC keys to top-level JSON (`quarkus.log.console.json.fields.*`); undeclared keys land nested under `mdc.*` at best (version-dependent) and won't match the Go side's top-level fields. Please confirm the JSON output actually contains these fields. (3) **Silent drop** — `putFields` (line 65) uses `i + 1 < length`, so a dangling odd key with no value is silently dropped; fine defensively, but it can hide a miswired call.

kichasov (commented):
**Follow-up review — quality/correctness beyond the P0 compile blockers.** The Java conversion output has systemic issues; the Go side is cleaner but has a design fragility. Counts across the aggregator diff: **~159 placeholder `argN` field keys** (arg0×98, arg1×37, arg2×13, arg3×7, arg4×3, arg5×1), **23 `StructuredLog` calls with a duplicated key**, **23 removed `error`/`warn` calls that passed a throwable**. Process: the Maven build is unverified in this PR (GitHub Packages 401) and controller tests fail (envtest), so the 545-file Java half is unproven; and the “632→0 `{}`” completion metric is misleading — 12 exception-mapper sites still template through a shared constant, so the counter reads 0 while the placeholders remain.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/controller/error/ForbiddenExceptionMapper.java
```diff
@@ -1,4 +1,5 @@
 package com.netcracker.cloud.dbaas.controller.error;
+import com.netcracker.cloud.dbaas.logging.StructuredLog;
```
vlsi, line 2: There's no point in adding unused imports. The shell script should remove the import if that was the only modified line.

dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/service/PasswordRotationService.java
```diff
@@ -161,26 +161,6 @@ PasswordChangeResponse performChangePassword(List<DatabaseRegistry> databasesFor
...
-    }
-
-    private Map<DbaasAdapter, Boolean> getAdaptersAndUserSupportedMap(List<DatabaseRegistry> databases) {
-        return databases
-                .stream()
-                .map(dbRegistry -> dBaaService.getAdapter(dbRegistry.getDatabase().getAdapterId()).orElseThrow(() ->
-                        new UnregisteredPhysicalDatabaseException("Adapter identifier: " + dbRegistry.getAdapterId())
-                ))
-                .distinct()
-                .collect(Collectors.toMap(adapter -> adapter, DbaasAdapter::isUsersSupported));
```
vlsi, line 184: Was this code removal intentional?

PR #575, opened 2026-07-10
chore: updating operator comments

TheHollowMonarch:


kichasov (commented):
Reviewed the full diff — comments and regenerated CRD templates only, no functional changes. The direction is right: most of the removed text was noise, and the shorter wording reads well. Three things before merge.

## Branch is behind the base

The diff is built against a pre-#599 `feat/operator-dev`: it still shows `HasBlockingResources`, the old `NewKindChecker` signature, and a NamespaceBinding without status. #599 rewrote exactly the files this PR touches (`namespacebinding_controller.go`, `checker.go`, `events.go`, `common_types.go`), so:

- merge the base and resolve carefully — this PR trims the `conditions` godoc block in `OperatorStatus`, and #599 just added the NamespaceBinding reasons to that same block; the trim must not drop them;
- rerun `make manifests generate sync-helm-crds` afterwards, or the `verify-generated` job will fail the PR.

## Two comments worth keeping

1. **`requestIDFromContext` (helpers.go)** — the new one-liner hides the contract: the function still panics when the context was not initialized. One line covers it: "panics if the context was not initialized".
2. **`handleAggregatorError` (helpers.go)** — the removed table (`401 → BackingOff; 400/403/409/410/422 → InvalidConfiguration; 5xx/network → BackingOff`) is the error-classification contract, mirrored in AGENTS.md, and it is not readable at a glance from the body (it hides behind `IsAuthError`/`IsSpecRejection`). Please keep the compact table.

## Minor, at your discretion

- `var _ runtime.Object = &dbaasv1.NamespaceBinding{}` lost its half-line explanation ("required for recorder.Eventf") and now looks cryptic on its own.
- Dropping the Step 9.1–9.4 numbering in `writeSecret` is fine since the per-branch race rationales survived, though the four-line sequence summary was more useful than the two-line replacement.
- The new `applySecretCredentials` comment ("duplicate target names would overwrite each other") is an improvement over the old one — it explains why.

After the base merge, regeneration, and the two restored comments this is good to merge.

kichasov:
**[P2]** `api/v1/common_types.go:56` (+ the same text regenerated into all 16 CRD descriptions): the condensed conditions godoc keeps the Ready reason catalog only for NamespaceBinding and drops it for every other kind, together with the "False on any error" rule.

`kubectl explain externaldatabase.status.conditions` now documents `BindingBlocked` and `BindingReleased`, but says nothing about `DatabaseRegistered`, `PolicyApplied`, or `DatabaseProvisioned`/`ProvisioningStarted` — the reasons a user of that very CRD greps for. The asymmetry is worse than either extreme: document all kinds or none. The new `BindingReleased` line also lost its distinguishing clause — "after its protection finalizer is removed" reads as any deletion; the point of the reason is that only other controllers' finalizers still keep the object alive (added after the #599 review specifically to capture that).

Suggestion — keep the condensation win but restore the contract; this stays at half the original size:

```go
// conditions represent the current state of the resource.
// Ready is True when the current generation was successfully processed;
// False on any error, with Reason carrying the error category. Success
// reasons: DatabaseRegistered (ExternalDatabase), PolicyApplied
// (DatabaseAccessPolicy), DatabaseProvisioned (InternalDatabase;
// ProvisioningStarted while the async operation runs), BindingRegistered
// (NamespaceBinding; BindingBlocked while deletion is deferred,
// BindingReleased once only other controllers' finalizers keep the object
// alive, OwnershipCheckError when listing blocking resources fails).
// Stalled=True marks a permanent error; the controller does not retry
// until the spec changes. Stalled=False covers successful, ongoing, and
// retriable states.
```

Needs `make manifests` + the helm CRD copy sync after the edit.

kichasov:
**[P3]** `internal/ownership/resolver.go` (package godoc) + `internal/controller/helpers.go` (`checkOwnership`): the condensed comments keep *what* the Unbound state does ("requeues slowly as a safety net") but drop *what it is a safety net against*, in both places at once.

The deleted text was the only documentation of a non-obvious design point: the slow requeue exists because the NamespaceBinding → workloads watch fan-out can lose its trigger on a transient LIST error, and the interval is deliberately long so permanently-unbound namespaces do not generate background churn. Without that, a future reader cannot tell why the requeue exists at all, why the interval must stay long, or whether it is safe to remove — which is exactly the kind of comment worth keeping when everything else gets condensed.

One clause in `checkOwnership` is enough, e.g.:

```go
// checkOwnership returns whether reconciliation should proceed for namespace.
// Unknown ownership requeues quickly; Unbound requeues at a long interval as a
// safety net against lost NamespaceBinding→workload fan-out triggers (the
// interval is long so permanently-unbound namespaces cause no churn); Foreign
// returns without requeue.
```

kichasov:
**[P4]** `internal/controller/externaldatabase_controller.go` (`SetupWithManager`): the condensed godoc says credential Secret changes are picked up "by periodic resync or by changing AnnotationRefresh", but drops *why there is no Secret watch in the first place*: the operator intentionally holds only namespaced Secret RBAC, no cluster-wide list/watch.

That line was the security rationale for the whole resync design. Without it, adding a Secret watch looks like an obvious improvement to a future reader, and the first review of such a change loses the one sentence that explains the trade-off. One parenthetical keeps it:

```go
// SetupWithManager registers watches for spec changes, refresh annotations, and
// NamespaceBinding fan-out. There is intentionally no Secret watch (the operator
// holds only namespaced Secret RBAC, no cluster-wide list/watch); credential
// Secret changes are picked up by periodic resync or by changing AnnotationRefresh.
```

kichasov:
**[P4]** `internal/controller/helpers.go` (`handleAggregatorError`): the rewritten mapping list switches the arrows from `→` to ASCII `->`, while the surviving comments elsewhere keep `→` (e.g. the rotation-poller note in `cmd/main.go`: "operator → aggregator"). Cosmetic, but the codebase now mixes both styles in the same package. Pick one — the remaining comments use `→`, so restoring it here is the smaller change.

PR #580, opened 2026-07-16
feat: Metrics to account for discrepancies between the registry and the actual 

TheHollowMonarch:


TheHollowMonarch:
will be handled in https://github.com/Netcracker/qubership-dbaas/pull/581

PR #586, opened 2026-07-21
fix: adding the finalizer permission for secret claim

TheHollowMonarch:
Added missing finalizers RBAC permission for **DatabaseSecretClaim** to fix **blockOwnerDeletion** errors when creating owned Secrets


dbaas-operator/helm-templates/dbaas-operator/templates/ClusterRole.yaml
```diff
@@ -33,6 +33,9 @@ rules:
   - apiGroups: [ dbaas.netcracker.com ]
     resources: [ databasesecretclaims/status ]
     verbs: [ get, update, patch ]
+  - apiGroups: [ dbaas.netcracker.com ]
```
kichasov, line 38: the PR adds databasesecretclaims/finalizers only to the Helm ClusterRole, but DatabaseSecretClaim has no code path that adds/
    removes a finalizer, and there is no kubebuilder RBAC marker for this resource. make manifests keeps config/rbac/role.yaml without this permission. If this permission is genuinely required, add the marker/generated RBAC too; otherwise this
    is an unnecessary cluster-wide RBAC expansion.

also check dev/k8s/operator.yaml

kichasov:
Verified this fix against a live cluster, and it turns out the bug is invisible to every test we have today. Details below, plus what it would take to make it visible.

## Why nothing catches this

`OwnerReferencesPermissionEnforcement` is not enabled by default, and kind does not enable it either. Our kind clusters run with:

```
--enable-admission-plugins=NodeRestriction
```

So the missing permission has no effect there, and envtest cannot catch it at all — those tests use an admin client, so the operator's ServiceAccount and its ClusterRole are never exercised. That is why `DatabaseSecretClaim` reaches `Succeeded` in the dev environment today despite the gap.

## Reproduction with the plugin on

I recreated the dev kind cluster with the plugin enabled and applied `dev/test-resources/dsc-success.yaml`.

**Without this PR's RBAC** — CR stuck in `Processing`, Secret never created, reconcile looping on:

```
secrets "dsc-success-secret" is forbidden: cannot set blockOwnerDeletion if an ownerReference
refers to a resource you can't set finalizers on
```

**With this PR's RBAC applied to the same cluster, nothing else changed** — CR reaches `Succeeded`, the Secret is created, and it carries:

```json
{
  "apiVersion": "dbaas.netcracker.com/v1",
  "kind": "DatabaseSecretClaim",
  "blockOwnerDeletion": true,
  "controller": true
}
```

So the fix is confirmed, and it is required on any cluster where the plugin is enabled: `ctrl.SetControllerReference` sets `blockOwnerDeletion: true`, and the admission plugin then demands `update` on the owner's `finalizers` subresource.

## Review notes on the change itself

- **Complete.** `SetControllerReference` is called only in `databasesecretclaim_controller.go` (lines 365 and 780). No other controller creates owned objects, so no other CR needs the same permission.
- **All three RBAC copies agree.** I reran `make manifests` on this branch: `config/rbac/role.yaml` matches the generated output byte for byte, so it was regenerated rather than hand-edited. The ClusterRole entry also sits in the file's per-CR grouping (resource, `/status`, `/finalizers`), which is the right place.
- **One suggestion.** The neighbouring blocks in `ClusterRole.yaml` each carry a comment, and the reason here is different from theirs: `namespacebindings/finalizers` is about changing `metadata.finalizers`, whereas this one is about an ownerReference. Without a note the next reader will assume the former. Something like:

  ```yaml
  # databasesecretclaims/finalizers: required because the operator sets an ownerReference
  # with blockOwnerDeletion on the Secret it creates (SetControllerReference).
  ```

## Enabling the plugin so this cannot regress

Worth doing separately, but recording it here since this PR is what surfaced it. Three places need the same patch.

`.github/workflows/integration-tests.yml`:

```yaml
      - name: Create kind config
        run: |
          cat <<EOF > kind-config.yaml
          kind: Cluster
          apiVersion: kind.x-k8s.io/v1alpha4
          kubeadmConfigPatches:
            - |
              kind: ClusterConfiguration
              apiServer:
                extraArgs:
                  # extraArgs replaces the flag rather than merging into it, so
                  # NodeRestriction — kind's own default — must be repeated here.
                  - name: enable-admission-plugins
                    value: NodeRestriction,OwnerReferencesPermissionEnforcement
          nodes:
            - role: control-plane
            - role: worker
            - role: worker
          EOF
```

`.github/workflows/integration-tests-sample-service.yml`: the same `kubeadmConfigPatches` block, keeping the existing `EOF_KIND` delimiter and the `region: database` label on the first worker.

`dbaas-operator/dev/kind-up.sh` currently creates the cluster with no config at all, so it needs one:

```bash
cat <<'EOF_KIND' > "${TMP_KIND_CONFIG}"
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    apiServer:
      extraArgs:
        - name: enable-admission-plugins
          value: NodeRestriction,OwnerReferencesPermissionEnforcement
EOF_KIND
kind create cluster --name "${KIND_CLUSTER}" --config "${TMP_KIND_CONFIG}"
```

Two things I checked rather than assumed:

- **The `v1beta4` list form is required.** Kubernetes 1.31+ uses kubeadm `v1beta4`, where `extraArgs` is a list of `{name, value}`. The map syntax found in most examples does not apply on the clusters we run (verified: server v1.36.1).
- **`extraArgs` replaces, it does not merge.** I recreated a cluster with only `OwnerReferencesPermissionEnforcement` in the patch and the resulting flag was exactly that — `NodeRestriction` silently disappeared. It is not in the Kubernetes default set, so it does not come back on its own.

## Third-party controllers under the plugin

The obvious risk is breaking something that is not ours, since the plugin applies cluster-wide. I checked the components the integration workflow deploys, on a three-node kind cluster matching the workflow's layout:

| Component | Result |
| --- | --- |
| MinIO operator + Tenant | StatefulSet and pod came up, no admission errors |
| patroni-core + pgskipper operator | operator and `pg-patroni-node1-0` running, master ready, no admission errors |

`dbaas-aggregator` creates no Kubernetes objects, and `patroni-services` is driven by the same pgskipper operator that already passed. So enabling the plugin looks safe for both workflows.

PR #587, opened 2026-07-21
chore(operator): enforce balancing-rule name and blank values in the CRD

kichasov:
## Why

The three balancing-rule CRDs had **no declarative validation at all** — every constraint lived in
the controller. A mistyped CR is therefore admitted, reconciled, moved to `InvalidConfiguration`,
and left in the cluster for someone to notice and delete. Two constraints are trivially expressible
in the schema, and one of them the controller does not fully cover either:

- **Singleton name.** Each balancing rule is a per-namespace singleton with a fixed name. The
  controller rejects a wrong name, but only after the object exists. `NamespaceBinding` already
  enforces its own singleton name with CEL — the balancing rules were the odd ones out.
- **Whitespace-only strings.** `MinLength=1` does not catch `" "` — it has length 1 and passes.
  The controller's `TrimSpace` checks caught it after admission; the schema now catches it before.

## What

- CEL rule on all three root types: `self.metadata.name == '<singleton>'`.
- `Pattern=\S` on `type`, `dbType`, `name`, `physicalDatabaseId`, and on the `microservices` /
  `namespaces` items.
- Regenerated `config/crd/bases` and synced the three Helm CRD templates.

**Controller-side validation is unchanged.** It covers what the schema cannot express — uniqueness
over the flattened `(type × microservice)` product, and the operator-namespace check that depends on
runtime config — and it still guards clusters running a stale CRD.

### Test change worth a look

One spec created a second `NamespaceBalancingRule` under a non-singleton name to stage a rule-name
collision. That shortcut no longer passes admission. The spec now creates the conflicting singleton
in **another namespace**, which is also the real scenario: aggregator rule names are global and each
namespace holds exactly one singleton. `validateNamespaceRuleGlobalConflicts` lists cluster-wide and
skips only itself, so the check it exercises is unaffected.

## How to verify

`make test` and `make lint` pass; two new specs assert the admission rejections. Verified against a
live API server in the kind dev environment:

```
$ kubectl apply -f <rule named my-custom-rules>
The NamespaceBalancingRule "my-custom-rules" is invalid: <nil>: Invalid value:
  NamespaceBalancingRule name must be 'namespace-balancing-rules'

$ kubectl apply -f <rule with name: "   ">
The NamespaceBalancingRule "namespace-balancing-rules" is invalid: spec.rules[0].name:
  Invalid value: "   ": spec.rules[0].name in body should match '\S'

$ kubectl apply -f <valid singleton>
namespacebalancingrule.dbaas.netcracker.com/namespace-balancing-rules created
```


TheHollowMonarch (approved):
The changes look good

PR #595, opened 2026-07-21
chore(ci): snapshot aggregator threads and PG locks during tests

kichasov:
> **Diagnostics — do not merge.** Stacked on #594. Meant to run until the freeze is caught, name the blocked frame, and
> then be replaced by the actual fix.

## Why

Some integration-test runs freeze every code path guarded by `DBAAS_REPOSITORIES_MUTEX`
([ServicesConfig.java:36](https://github.com/Netcracker/qubership-dbaas/blob/feat/operator-dev/dbaas/dbaas-aggregator/src/main/java/com/netcracker/cloud/dbaas/config/ServicesConfig.java#L36)),
a single static monitor shared by logical-database persistence, the H2 cache reload, and the Postgres table listeners.

In run [29850808300](https://github.com/Netcracker/qubership-dbaas/actions/runs/29850808300) the aggregator logged its
last `External logical database was saved`, `New database was created`, and `finished reload in databaseregistry` at
17:23:44. For the next 26 minutes it accepted 32 registration requests, logged `Skip creation` for each, and answered
none — while unrelated endpoints (Blue/Green, physical database registration) kept serving normally. The last thread to
make progress was `backups-pool-10-thread-9`:

```text
17:23:47.045 [DeletionService] Successfully dropped logical database 'operator-autotests-0ac8149d-…'
                              ← 26 minutes of silence
17:49:37.047 [DeletionService] Error happened during dropping registry 808107b8-… [Error Occurred After Shutdown]
17:49:37.051 [DeletionService] Register deletion error … of database operator-autotests-0ac8149d-…
```

Same thread, same database, and between those two log lines sits exactly one call —
`DatabaseRegistryDbaasRepository.delete`, which holds the mutex across both a Postgres delete and an H2 delete. The
thread came back only when the pod shut down. That is what makes the `OperatorIT` and `BlueGreenV1IT` failures look like
flakes: the operator's aggregator client gives up after 30s and retries, so the freeze surfaces as scattered timeouts.

Pod logs cannot go further: they show which call was entered, not which frame is parked.

## What

A background watchdog that runs for the whole test step and, once a minute:

- takes a JVM thread dump of every running aggregator pod (`jcmd Thread.print -l`, falling back to `SIGQUIT` when
  `jcmd` is missing, which routes the dump to the pod stdout Fluent Bit already collects);
- snapshots `pg_stat_activity` with `pg_blocking_pids`, transaction age, and wait events.

Both sides are needed: a thread parked on a Postgres row lock is an ordinary socket read in a thread dump, so only the
database snapshot distinguishes a JVM monitor from a lock held by another backend. Snapshots upload as the
`aggregator-stall-snapshots` artifact.

The script never fails the job — no `set -e`, every `kubectl` call guarded, missing pods skipped.

## Status

First run, [29861265169](https://github.com/Netcracker/qubership-dbaas/actions/runs/29861265169), was green, so the
freeze did not reproduce. The collector itself works end to end:

- the image ships no `jcmd`, so the fallback ran; PID 1 is the JVM, and 36 thread dumps landed in the pod stdout
  artifact;
- 34 `pg_stat_activity` snapshots landed in `aggregator-stall-snapshots`;
- during the run the aggregator was genuinely idle — every dump shows zero `BLOCKED` threads and no active Postgres
  backends, which is the baseline a frozen run has to be compared against.

Re-run until a red run is captured.

## Follow-up

The fix itself is a separate change. Holding a JVM-wide monitor across JDBC and JTA work has no timeout and no fairness,
so a single blocked holder wedges the aggregator until restart. Related: `safeDeleteAndFlushDatabaseRegistry` is
annotated `@Transactional(REQUIRES_NEW)` but invoked through `this`, so no interceptor applies and it runs in the
caller's transaction.

### DIFF
--- a/.github/scripts/watch-aggregator-stalls.sh
+++ b/.github/scripts/watch-aggregator-stalls.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+# Periodic snapshots of what the aggregator and its PostgreSQL are blocked on.
+#
+# Aimed at the stall where every code path guarded by DBAAS_REPOSITORIES_MUTEX
+# freezes: logical database writes and H2 cache reloads stop, external database
+# registrations pile up unanswered, and the aggregator only recovers when the pod
+# is replaced. Log lines alone cannot name the blocked frame, so each round takes
+# a JVM thread dump and a PostgreSQL lock snapshot at the same moment.
+#
+# Runs in the background for the whole test step, so it must NEVER fail the job:
+# no `set -e`, every kubectl call guarded, unavailable pods skipped.
+set -uo pipefail
+
+OUT_DIR="${OUT_DIR:-./thread-dumps}"
+INTERVAL="${INTERVAL:-60}"
+AGG_NS="${AGG_NS:-dbaas}"
+AGG_DEPLOY="${AGG_DEPLOY:-dbaas-aggregator}"
+PG_NS="${PG_NS:-postgres}"
+PG_POD="${PG_POD:-pg-patroni-node1-0}"
+
+mkdir -p "$OUT_DIR"
+
+# SIGQUIT prints the dump to the JVM's stdout, which Fluent Bit already collects.
+# jcmd returns it on our own stdout instead, which keeps the dumps timestamped and
+# separate from a multi-megabyte pod log, so try jcmd first.
+dump_threads() {
+  local pod=$1 stamp=$2
+  local out="$OUT_DIR/${pod}-${stamp}.threads.txt"
+
+  # shellcheck disable=SC2016  # the body runs in the pod's shell, not this one
+  kubectl -n "$AGG_NS" exec "$pod" -c "$AGG_DEPLOY" -- sh -c '
+    pid=""
+    for proc in /proc/[0-9]*; do
+      [ -r "$proc/cmdline" ] || continue
+      case "$(tr "\0" " " < "$proc/cmdline")" in
+        *java*) pid="${proc#/proc/}"; break ;;
+      esac
+    done
+    [ -n "$pid" ] || { echo "no java process found"; exit 1; }
+
+    if command -v jcmd >/dev/null 2>&1; then
+      jcmd "$pid" Thread.print -l
+    else
+      echo "jcmd unavailable, sent SIGQUIT to pid $pid — dump is in the pod stdout"
+      kill -3 "$pid"
+    fi
+  ' > "$out" 2>&1
+
+  if [ ! -s "$out" ]; then
+    rm -f "$out"
+  fi
+}
+
+# A thread parked on a PostgreSQL row lock looks like an ordinary socket read in the
+# thread dump. pg_stat_activity names the blocker, so capture both sides.
+dump_pg_locks() {
+  local stamp=$1
+  kubectl -n "$PG_NS" exec "$PG_POD" -- psql -U postgres -d postgres -X -P pager=off -c "
+    SELECT now() AS captured_at, a.pid, a.state, a.wait_event_type, a.wait_event,
+           age(clock_timestamp(), a.xact_start) AS xact_age,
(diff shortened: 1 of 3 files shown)


kichasov:
## Reproduced — cross-layer deadlock confirmed

Six parallel re-runs on this branch; run [29868895200](https://github.com/Netcracker/qubership-dbaas/actions/runs/29868895200) froze (59 min, 18 failures). The watchdog caught it from both sides.

### JVM side — one dump, 56 threads `BLOCKED` on the same monitor

All 56 wait on monitor `0xfa764078` — the `DBAAS_REPOSITORIES_MUTEX` static `Object`. By entry point: 32 `saveExternalDatabase`, 14 `createDatabaseFromRequest`, 4 `dropRegistrySafe`, 2 `PeriodicH2Reload`.

The holder, `pool-5-thread-5`, is `RUNNABLE` but parked in a Postgres socket read, holding the monitor:

```text
sun.nio.ch.Net.poll(Native Method)
...
org.postgresql.jdbc.PgPreparedStatement.executeUpdate      ← UPDATE database ...
org.hibernate.internal.SessionImpl.flush
DatabaseRegistryDbaasRepositoryImpl.save(...:321)          ← inside synchronized(getMutex())
DatabaseRegistryDbaasRepositoryImpl.saveInternalDatabase(...:134)
DBaaService.encryptAndSaveDatabaseEntity(...:394)
AggregatedDatabaseAdministrationService.createNewDatabase(...:302)
BlueGreenService.createOrUpdateDatabaseWarmup(...:309)
NewDatabaseTask.executeTask(...:35)
- locked <0x00000000fa764078> (a java.lang.Object)
```

`elapsed=2291s`, `cpu=55ms` — parked ~38 minutes, not spinning.

### Postgres side — same second, `pg_stat_activity`

```text
pid | state               | wait_event    | xact_age | blocked_by | query
226 | idle in transaction | ClientRead    | 24:11    | {}         | update database_state_info set database_state=$1,... where id=$5
240 | active              | transactionid | 24:11    | {226}      | update database set adapter_id=$1,... 
```

- **pid 240** is `pool-5-thread-5` — holds the JVM monitor, its `UPDATE database` blocked on a row lock held by 226.
- **pid 226** ran an `update database_state_info` (holds that row lock), then went `idle in transaction`: its owning thread is one of the 56 blocked entering the monitor, so it never issues its next statement.

### The cycle

```text
thread owning pid 226 ──wants──▶ DBAAS_REPOSITORIES_MUTEX ──held by──▶ pool-5-thread-5 (pid 240)
        ▲                                                                      │
        └──────────────── PG row lock on database/database_state_info ◀────wants┘
```

`synchronized (DBAAS_REPOSITORIES_MUTEX)` wraps a Hibernate `flush()` that takes Postgres row locks, while another thread takes a Postgres row lock *before* trying to enter the same monitor — opposite acquisition order for `{DB row lock, JVM monitor}`. Postgres can't detect it (226 is idle, not waiting in PG); the JVM can't either (the monitor holder is `RUNNABLE` in native `poll`). It never resolves without a pod restart, which is exactly the 26-minute silence seen in run 29850808300.

### Fix direction (separate PR)

1. Don't hold the process-wide monitor across the Hibernate `flush()` / JDBC write. Narrow it to the in-memory H2 cache mutation it was meant to serialize, so the DB write happens outside the monitor and the cross-layer cycle can't form.
2. Safety net: a `ReentrantLock.tryLock(timeout)` in place of `synchronized`, plus a Postgres `statement_timeout` / `lock_timeout`, so any residual contention surfaces as a bounded, logged error instead of a permanent freeze.
3. Separately, no transaction should stay `idle in transaction` across a lock acquisition — worth auditing the `@Transactional` boundary that leaves pid 226 open.

kichasov:
Diagnostic PR — served its purpose, closing.

PR #604, opened 2026-07-24
feat: improve DBaaS operator dashboard

TheHollowMonarch:
Updates the DBaaS Operator Grafana dashboard to support multiple operator instances and DR clusters.
- Generates a namespace-specific dashboard UID to prevent collisions.
- Adds namespace and cluster selectors to all queries.
- Improves panel layout, tags, descriptions, and default time settings.
- Documents dashboard behavior, including conditionally empty panels.

### DIFF
--- a/dbaas-operator/docs/monitoring/DBaaS Operator Metrics.md
+++ b/dbaas-operator/docs/monitoring/DBaaS Operator Metrics.md
@@ -158,8 +158,10 @@ The dashboard exposes three variables:
 
 Every panel query is scoped by `cluster=~"$cluster"` and `namespace=~"$namespace"`. This allows one
 provisioned dashboard to switch between operator instances without changing its UID. The default
-range is the last 30 minutes and automatic refresh is disabled; refresh manually when testing a
-short-lived state such as deletion.
+range is the last 30 minutes and the dashboard refreshes every minute. This operational dashboard
+uses the monitoring design guide's explicit auto-refresh exception so that reconciliation lag,
+active errors, deletion state, and time-based health metrics stay current. With metrics scraped
+every 30 seconds, this issues one dashboard query refresh per two scrape cycles.
 
 **To open it:** in Grafana, open the **DBaaS Operator** dashboard (the name may differ per
 installation), select the **datasource**, **cluster**, and **DBaaS Operator Namespace**.
--- a/dbaas-operator/docs/monitoring/DBaaS Operator Metrics.md
+++ b/dbaas-operator/docs/monitoring/DBaaS Operator Metrics.md
@@ -166,7 +166,7 @@ every 30 seconds, this issues one dashboard query refresh per two scrape cycles.
 **To open it:** in Grafana, open the **DBaaS Operator** dashboard (the name may differ per
 installation), select the **datasource**, **cluster**, and **DBaaS Operator Namespace**.
 The dashboard is organized into rows that mirror the metric groups above. `CR Health Overview`
-is expanded by default as the landing overview; all detailed rows are collapsed until needed.
+appears first and is expanded by default; all detailed rows are collapsed until needed.
 
 ### Credentials & Secret Resolution
 
--- a/dbaas-operator/docs/monitoring/DBaaS Operator Metrics.md
+++ b/dbaas-operator/docs/monitoring/DBaaS Operator Metrics.md
@@ -143,16 +143,18 @@ title **"DBaaS Operator"**) from `templates/Dashboard.yaml`, gated on
 `DBAAS_OPERATOR_ENABLED && MONITORING_ENABLED` (chart defaults: `DBAAS_OPERATOR_ENABLED: false`,
 `MONITORING_ENABLED: true` — so it ships once the operator is enabled) and reconciled by the
 **Grafana Operator**. Its UID follows the monitoring convention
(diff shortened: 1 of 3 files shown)


kichasov:
**[P2]** `dashboards/dbaas-operator-dashboard.json` (the `namespace` variable) + `templates/Dashboard.yaml`: the `namespace` variable has no `current` default, so a freshly opened dashboard shows an arbitrary namespace instead of its own.

Each install provisions its own dashboard (the UID is namespace-specific now), but on first load Grafana picks the alphabetically first value the variable query returns. With several operator instances in one cluster, the dashboard provisioned by install A can open on install B's metrics — easy to misread during an incident, and the install-time pre-scoping this PR removes used to guarantee the opposite.

The chart already knows `.Values.NAMESPACE`, so the default can be substituted the same way as the UID — a second placeholder in the JSON:

```json
"current": {
    "selected": true,
    "text": "namespace-default-placeholder",
    "value": "namespace-default-placeholder"
},
```

and in `Dashboard.yaml`:

```yaml
json: {{ .Files.Get "dashboards/dbaas-operator-dashboard.json" | toJson | replace "dashboard-uid-placeholder" $dashboardUID | replace "namespace-default-placeholder" .Values.NAMESPACE }}
```

The dashboard then opens scoped to its own operator while staying switchable.

kichasov:
**[P4]** `dashboards/dbaas-operator-dashboard.json`: the PR removes `"refresh": "30s"`, so the dashboard no longer auto-refreshes at all.

For an operational dashboard that seems backwards: during an incident the panels silently go stale unless the operator on duty remembers to refresh. The doc's rationale ("refresh manually when testing a short-lived state such as deletion") also reads inverted — a short-lived state is exactly what a periodic refresh helps catch. If the intent was to cut datasource load, 1m would do that while keeping the dashboard live; if disabling was deliberate for another reason, the doc sentence should state it.

kichasov:
**[P4]** `docs/monitoring/DBaaS Operator Metrics.md`, wording nits:

- "truncates the namespace portion to keep the Grafana 40-character limit" → "to stay within Grafana's 40-character UID limit" (you keep a promise, you stay within a limit; and it is the UID limit specifically).
- The new variable list uses plain hyphens as separators ("**`datasource`** - Prometheus, ..."), while the surrounding document uses em-dashes for that role (the text this PR replaces did too: "one variable — **`datasource`** —"). Consistency either way, but em-dash matches the rest of the file.

kichasov:
**[P2]** `templates/Dashboard.yaml`: the truncation-based UID defeats the PR's own goal — two namespaces that share their first 25 characters render the *same* UID.

`.Values.NAMESPACE | trunc 25` keeps only the prefix, so e.g. `dbaas-operator-production-east` and `dbaas-operator-production-west` both become `dbaas-operator-production-dbaas-operator` — one dashboard overwrites the other, which is exactly the collision this PR set out to prevent (verified: Helm renders identical UIDs for two such namespaces). Long namespace names are also the realistic case here, since the ones that need truncation are the ones that collide.

Keep the readable prefix but make the UID unique by appending a short hash of the full namespace; the budget still fits Grafana's 40-character UID limit exactly (18 + 1 + 6 + 15):

```yaml
{{- $nsHash := sha256sum .Values.NAMESPACE | trunc 6 }}
{{- $dashboardNamespace := .Values.NAMESPACE | trunc 18 | trimSuffix "-" }}
{{- $dashboardUID := printf "%s-%s-dbaas-operator" $dashboardNamespace $nsHash }}
```

Short namespaces stay recognizable (`prod-a1b2c3-dbaas-operator`), and any two distinct namespaces get distinct UIDs. The doc sentence about the UID convention needs the same update.

TheHollowMonarch:
> **[P4]** `dashboards/dbaas-operator-dashboard.json`: the PR removes `"refresh": "30s"`, so the dashboard no longer auto-refreshes at all.
> 
> For an operational dashboard that seems backwards: during an incident the panels silently go stale unless the operator on duty remembers to refresh. The doc's rationale ("refresh manually when testing a short-lived state such as deletion") also reads inverted — a short-lived state is exactly what a periodic refresh helps catch. If the intent was to cut datasource load, 1m would do that while keeping the dashboard live; if disabling was deliberate for another reason, the doc sentence should state it.

The design document mentions setting no automatic refresh other than situations which really need it. I feel that adding a 1minute refresh is fine (not too many queries, fits in well with the scraping of metrics). Have re-added it with a refresh of 1minute.

kichasov:
Superseded by #610, which carries these changes onto `feat/operator-dev` with your authorship preserved.

Retargeting this PR in place would not have worked: it is cut from `main`, and because #542 reached `main` as a **squash** merge, this branch and `feat/operator-dev` share only the 24 July merge base. Changing the base alone turned the diff into 187 files / +13981 lines — `feat/operator-dev`'s own work coming back at it through the squash commit. #610 is branched from `feat/operator-dev` instead, so it shows the intended three files.

Nothing was dropped in the port: the dashboard JSON and `Dashboard.yaml` were taken verbatim, and the only conflict — the "To open it" paragraph, also rewritten by #608/#609 — was resolved in favor of the wording here, since it is the one describing the new `cluster` and `namespace` variables.

Two review nits are still open and noted in #610 rather than silently changed: the dashboard title is still the fixed `DBaaS Operator` although the UID is now per-namespace, and the `datasource` variable is labelled **Cloud** in the UI while the doc calls it "datasource".

PR #611, opened 2026-07-27
fix: dashboard review followups

TheHollowMonarch:


kichasov:
**[P3]** The auto-refresh rationale now contradicts the one this branch shipped two commits ago. `Dashboard.yaml` in #610 said the dashboard "intentionally refreshes every minute … the monitoring design guide permits auto-refresh when justified"; the comment here says auto-refresh "can hinder issue investigation". Both statements cannot be the operative rule, and as written the new one argues against auto-refresh on *any* operational dashboard, which retroactively empties the exception cited earlier.

Keeping it disabled is a fine call — but the doc should not read as "live data is unavailable". Grafana's own refresh picker solves the stated problem better than a hard-off default: an engineer inspecting evidence pauses it, everyone else gets a moving view. One sentence covers it:

> Automatic refresh is off by default so that data does not shift while you are inspecting an incident. Set an interval in Grafana's refresh picker when you want the view to follow live data.

Related: with `now-30m` and no refresh, a dashboard left open keeps showing an ever-older 30-minute window. Worth stating explicitly next to the range, since the previous revision's 1m refresh made that a non-issue.

kichasov:
**[P3]** `dbaas-operator-dashboard.json` bumps `"version"` by hand (7 → 9). In a provisioned dashboard that field is Grafana's internal optimistic-concurrency counter, not a revision marker for the artifact: Grafana increments it itself on every UI save. Hand-setting it buys nothing, and it can mislead — if anyone edits the dashboard in the UI, the stored counter moves past the value in Git, and the next rollout ships a *lower* number, which reads like a rollback in Grafana's version history.

Cleaner to drop the key from the shipped JSON entirely; Grafana assigns it. If a revision marker is genuinely wanted, an annotation on the `GrafanaDashboard` CR (alongside the existing `deployment.netcracker.com/sessionId`) carries it without colliding with Grafana's own bookkeeping.

PR #634, opened 2026-08-06
fix: changing the value of settings from strings to apiextensionv1 json

TheHollowMonarch:
## Why

`InternalDatabase.spec.settings` only accepted string values, preventing structured settings such as
`pgExtensions: ["vector"]`.

## What

Allow settings values to contain arbitrary JSON, update the migration package, and add array and nested-object tests.

Existing string-valued CRs remain compatible. During upgrades, apply the updated CRD before starting the new operator
binary. The standard Helm upgrade path is safe because the chart ships both together.


kichasov (commented):
Reviewed the change — the type migration is done right and the test coverage is exemplary. Verified against the branch:

- `apiextensionsv1.JSON` is the canonical choice for arbitrary JSON in CRDs, and the generated schema (`x-kubernetes-preserve-unknown-fields` under `additionalProperties`) matches controller-gen output; both CRD copies (`config/crd/bases` and helm) are in sync.
- Backward compatibility is clean: existing string-valued CRs stay valid at admission, decode into `JSON{Raw: "\"...\""}`, and marshal to byte-identical wire JSON — no behavioral change for current users. The controller needed no changes (`buildPayload` passes `spec.Settings` straight through).
- `DeepCopyInto` is correct (`*val.DeepCopy()` per entry).
- The test pyramid is great: the envtest asserts the exact apply payload (string + array), and the new `OperatorIT` case proves the full path down to a live PostgreSQL — `pgExtensions: [uuid-ossp]`, `Succeeded/DatabaseProvisioned`, settings equality via get-by-classifier, and an actual `pg_extension` row check.

Remarks:

1. **[P2] The sibling APM package still assumes the old contract.** `migrate-core-operator-dbaas-declarations` keeps the string-map assumption: `references/mapping.md:46` ("flag non-string values if CRD requires string values"), `mapping.md:104` ("…when the target CRD schema is `map[string]string`"), and `scripts/convert_dbaas_crs.py:258` (the "has non-string settings…" warning). After this PR those warnings are stale for exactly the case this PR legalizes — legacy declarations with structured settings. Please update the wording, soften/drop the converter warning, and bump that package's version — either here or as an explicit follow-up.
2. **[P3] Empty PR description.** A schema change deserves a compatibility statement: existing string-valued CRs are unaffected; on upgrade the CRD must apply before the new binary (the helm chart ships them together, so the standard path is safe).
3. **[P3] No nested-object case in the envtest.** String and array are covered; adding `{"nested": {"a": 1}}` is one line and closes the deepest `preserve-unknown-fields` case.
4. **Observation, no action needed.** `x-kubernetes-preserve-unknown-fields` accepts arbitrarily large nested structures with no size guard — by design (pass-through validated by the adapter), just noting it.

Verdict: approve once the P2 is addressed here or tracked as a follow-up. No security or performance concerns.

kichasov (approved):
All review remarks are addressed:

- **[P2]** The `migrate-core-operator-dbaas-declarations` package is fully aligned: `SKILL.md` and both `mapping.md` spots now say "preserve JSON values verbatim", the `convert_dbaas_crs.py` non-string warning is removed (the non-object check is correctly kept), `apm.yml` is bumped to 1.1.0, and `examples.md` even gained a `pgExtensions` sample.
- **[P3]** The PR description now carries the Why/What and the compatibility statement (existing string-valued CRs unaffected; CRD applies before the binary; the standard Helm path ships both together).
- **[P3]** The envtest covers a nested object with the correct `float64` assertion.

Re-checked the package for stale `map[string]string` wording — none remains. Approving.

kichasov:
A follow-up finding, verified with a live repro on this branch: the `migrate-core-operator-dbaas-declarations` converter copies `settings` values verbatim and only checks that `settings` itself is an object (`convert_dbaas_crs.py:245-252`). "Arbitrary JSON" is narrower than "arbitrary YAML/Python value", and non-JSON values now pass through silently.

**Repro 1 — YAML input.** A legacy declaration with

```yaml
settings:
  encoding: UTF8
  timeout: .nan
  limit: .inf
```

converts with exit 0, zero settings-related warnings, and produces

```yaml
spec:
  settings:
    encoding: UTF8
    timeout: .nan
    limit: .inf
```

**Repro 2 — JSON input.** Line 90 calls bare `json.loads(text)`, which by default accepts the formally invalid `{"timeout": NaN}` (`parse_constant`); the converter emits the same `timeout: .nan` with exit 0.

**Downstream failure.** The generated manifest breaks before it even reaches the API server, in kubectl's client-side YAML-to-JSON conversion:

```text
$ kubectl create --dry-run=client -f out.yaml
error: error parsing out.yaml: error converting YAML to JSON: json: unsupported value: +Inf
```

The error names neither the resource nor the field, so the user gets a cryptic apply-time failure instead of a converter warning pointing at `settings.timeout`.

**Suggested fix.** This PR already ships the right check in the sibling package — `is_json_value()` in `dbaas-mounted-secret-migration/scripts/validate_generated.py` (recursive, `math.isfinite`, string keys) — and its `testing.md` explicitly demands negative fixtures for non-finite numbers and YAML-only values. Mirror it in `convert_dbaas_crs.py`:

- recursively validate `settings` values; reject non-finite numbers and non-string keys in nested objects (this also catches other YAML-only types such as timestamps and `!!binary`);
- emit a warning that names the resource and the full key path (`settings.timeout`), and do not present the result as ready to apply;
- add two negative tests — YAML `.nan` and JSON `NaN` (the latter also suggests calling `json.loads` with a `parse_constant` that rejects the nonstandard constants);
- keep valid strings, numbers, booleans, `null`, arrays, and nested objects passing through unchanged.

kichasov:
Verified fixed at `2f7a6f80` — closing my finding above.

- Ran the three new tests locally: all pass. The negative YAML case covers `.nan`, a YAML date, `!!binary`, and a non-string nested key, asserting full paths; the JSON case proves parse-time rejection of the bare `NaN`; the positive case round-trips every valid JSON type.
- Reran my original repro against the fixed converter:

  ```text
  ERROR: InternalDatabase legacy-db has invalid JSON value at settings.timeout: non-finite numbers are not valid JSON
  ERROR: InternalDatabase legacy-db has invalid JSON value at settings.limit: non-finite numbers are not valid JSON
  Conversion failed with 2 invalid setting value(s); the output file was not written
  exit=1
  ```

  Exactly the diagnostics that were missing: resource, full key path, reason — and no poisoned output file. Going error-plus-no-write instead of warning-plus-write is stricter than I suggested and the right call.

One follow-up suggestion, not blocking: the new `tests/test_convert_dbaas_crs.py` is not wired into any CI workflow, so it only runs manually. A `python -m unittest discover` step over `agent-packages/*/tests` in an existing workflow would keep it from going stale; the sibling package's `validate_generated.py` could adopt the same `tests/` pattern later.

My approval stands.

dbaas/dbaas-integration-tests/src/test/java/com/netcracker/it/dbaas/test/declarative/OperatorIT.java
```diff
@@ -754,6 +754,33 @@ void testInternalDatabaseValidMinimalSpec() throws IOException {
                     helperV3.getDatabaseByClassifierAsPOJO(helperV3.getClusterDbaAuthorization(), new ClassifierBuilder().ms(microserviceName).ns(NAMESPACE).build(), NAMESPACE, "postgresql", 200);
                 }
 
+                @Test
+                void testInternalDatabaseWithPostgresqlExtensions() throws IOException, SQLException {
```
ArkuNC: Can we check that logic in some existing test? Like in testInternalDatabaseCanUpdateSettings. To speedup tests.

ArkuNC (changes requested):

ArkuNC (approved):

PR #635, opened 2026-08-07
docs(operator): correct comments that misstate what the code does

vlsi:
## Why

Comments in `dbaas-operator` had drifted from the code they describe. Some were wrong in ways a reader would act on: a CRD description told users a field travels in the request URL path when the operator sends it in the body; `EventReasonAggregatorRejected` claimed it covers "a 4xx error other than 401" when it is chosen for five specific codes and for a failed async operation; `OwnershipResolver.SetOwner` named its argument "the binding's location" when both call sites pass the operator's own namespace. Others were structurally unusable: `status.conditions` listed the Ready reasons for four of the eight kinds and had never been extended, so a reader who did not find their reason concluded the wrong thing.

None of this is caught by the build. A comment that names a symbol which no longer exists still compiles, a Go doc link that resolves to nothing renders as literal text, and a CRD description ships to users through `kubectl explain` with no review step of its own.

## What

Four commits, one per area, comments only:

| Commit | Area |
|---|---|
| `d36bb97` | `internal/client`, `internal/poller`, `internal/ownership` |
| `7f5968d` | `api/v1` + the CRDs regenerated from it |
| `a3848e5` | `internal/controller` |
| `faa7df5` | `cmd`, `dev/aggregator-mock` |

Corrections outnumber additions in the parts that matter. A representative sample of what was wrong:

- `spec.dbName` said it is included in the request URL path; it is sent in the registration request body.
- `MockRule` said the ConfigMap is the single source of truth for every response; the mock hardcodes two, and three of the six files it loads hold a different shape entirely.
- `buildPayload` said the entire spec is forwarded as-is; `toWireSpec` reshapes it and defaults `classifier.namespace`.
- `preflightValidate` enumerated the checks it runs and omitted the `extraKeys` collision check, which makes the list worse than no list.
- `conditions.go` said `Stalled=False` meant a transient error; it also covers a successful and an in-progress reconcile.
- The `DatabaseAccessPolicy` watch comments called the CRs "DbPolicies", which is the aggregator's wire `subKind` rather than the kind of the CR.
- `updateOwnedSecret` said a metadata-only rewrite reports `SecretCreated`; that path sets `SecretUpToDate`.

Contracts that were absent rather than wrong now have a home: `RotationPoller.Interval` panics the ticker when non-positive; `Start` never returns an error, so no poll failure reaches the manager; `IsMyNamespace` writes its result back to the cache and cannot distinguish `Foreign` from `Unbound`; `CompositeChecker.Add` takes no lock; the mock's four loaders call `log.Fatalf` on a parse error, so a malformed file kills the process.

`status.conditions` now names all 23 Reason values grouped by the kind that reports them, verified against `events.go` in the same change, with the four condition-only ones called out. `api/v1` edits are regenerated into `config/crd/bases` and the chart's CRD templates in the same commit, so `verify-generated` stays green.

Cut throughout: accounts of how the code came to be this way, caller lists that find-usages answers (two of which were incomplete), and rationale for why a helper exists at all.

## How to verify

Nothing executable changed. That is checkable rather than asserted — strip the comments from every file in this branch and what remains is byte-identical to the base:

```
make -C dbaas-operator lint test-unit
make -C dbaas-operator manifests generate sync-helm-crds && git status --porcelain   # must be empty
```

`make lint` is red at base on two findings that predate this branch and are untouched by it: `SA1019` in `api/v1/groupversion_info.go` and `unparam` in `dev/aggregator-mock/main_test.go`. Gate on "no new findings" rather than on a clean run.

`make test` is not the check to run here — it needs envtest binaries. `make test-unit` covers everything this branch could affect.

### DIFF
--- a/dbaas-operator/api/v1/balancingrule_types.go
+++ b/dbaas-operator/api/v1/balancingrule_types.go
@@ -116,6 +116,7 @@ type MicroserviceBalancingRuleStatus struct {
 
 // MicroserviceBalancingRule is the Schema for the microservicebalancingrules API.
 // It declares a physical database placement rule for specific microservices in a namespace.
+// The object must be named microservice-balancing-rules, so a namespace can hold only one.
 type MicroserviceBalancingRule struct {
 	metav1.TypeMeta `json:",inline"`
 
@@ -222,6 +223,7 @@ type NamespaceBalancingRuleStatus struct {
 
 // NamespaceBalancingRule is the Schema for the namespacebalancingrules API.
 // It declares a namespace-level physical database placement rule.
+// The object must be named namespace-balancing-rules, so a namespace can hold only one.
 type NamespaceBalancingRule struct {
 	metav1.TypeMeta `json:",inline"`
 
@@ -321,6 +323,9 @@ type PermanentBalancingRuleStatus struct {
 
 // PermanentBalancingRule is the Schema for the permanentbalancingrules API.
 // It declares a permanent namespace-level physical database placement rule.
+// The object must be named permanent-balancing-rules and must live in the
+// namespace the operator runs in. Any other name or namespace is reported as an
+// invalid configuration, and the rules are not applied.
 type PermanentBalancingRule struct {
 	metav1.TypeMeta `json:",inline"`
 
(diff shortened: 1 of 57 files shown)


vlsi:
Rebased PR on top of `next` branch.

PR #636, opened 2026-08-07
fix: propagate request id to the aggregator

TheHollowMonarch:
## Fix

Previously, the operator generated an `X-Request-Id` for each reconcile and exposed it through logs and `status.lastRequestId`, but the ID was not included in requests sent to dbaas-aggregator.

This change:

- propagates the request ID through the shared Resty client for every aggregator endpoint;
- generates a fresh request ID for every reconcile and rotation-poller iteration;
- centralizes provider registration and request-ID initialization;
- validates the request context before sending the request;
- classifies missing or invalid request context as permanent `InvalidConfiguration` / `Stalled=True` instead of a retryable network failure;
- reports these failures as `configuration_error` in aggregator request metrics;
- adds regression coverage for startup registration, HTTP header propagation, controller status handling, InternalDatabase polling, and the rotation poller.

The request ID sent to the aggregator now matches the ID associated with the corresponding operator operation, enabling correlation between operator and aggregator logs.

## Scope

Each poll and reconcile propagates its own request ID. Linking a poll request to the multiple reconciles it asynchronously triggers through Kubernetes requires a separate parent-correlation design and is outside the scope of this PR.

### DIFF
--- a/dbaas-operator/internal/client/aggregator_client.go
+++ b/dbaas-operator/internal/client/aggregator_client.go
@@ -30,7 +30,10 @@ import (
 
 	"github.com/go-resty/resty/v2"
 	"github.com/netcracker/qubership-core-lib-go-error-handling/v3/tmf"
+	"github.com/netcracker/qubership-core-lib-go/v3/context-propagation/baseproviders/xrequestid"
+	"github.com/netcracker/qubership-core-lib-go/v3/context-propagation/ctxhelper"
 	"github.com/netcracker/qubership-core-lib-go/v3/security/tokensource"
+	"github.com/netcracker/qubership-dbaas/dbaas-operator/internal/requestcontext"
 )
 
 const defaultTimeout = 30 * time.Second
@@ -113,6 +116,20 @@ func newClient(baseURL string, getToken func(ctx context.Context) (string, error
 		// so the warning is pure noise logged on every call (incl. the rotation poll loop).
 		SetDisableWarn(true).
 		OnBeforeRequest(func(_ *resty.Client, r *resty.Request) error {
+			if _, err := requestcontext.RequestID(r.Context()); err != nil {
+				return &RequestContextError{Cause: err}
+			}
+			if err := ctxhelper.AddSerializableContextData(r.Context(), func(name, value string) {
+				r.SetHeader(name, value)
+			}); err != nil {
+				return &RequestContextError{Cause: fmt.Errorf("serialize request context: %w", err)}
+			}
+			if r.Header.Get(xrequestid.X_REQUEST_ID_HEADER_NAME) == "" {
+				return &RequestContextError{Cause: fmt.Errorf(
+					"serialization did not add required header %s; verify that request-ID propagation is not restricted",
+					xrequestid.X_REQUEST_ID_HEADER_NAME)}
+			}
+
 			// M2M mode — fetch a fresh dbaas-audience token per request.
 			if c.getToken != nil {
 				token, err := c.getToken(r.Context())
--- a/dbaas-operator/internal/poller/rotation_poller.go
+++ b/dbaas-operator/internal/poller/rotation_poller.go
@@ -23,6 +23,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	aggregatorclient "github.com/netcracker/qubership-dbaas/dbaas-operator/internal/client"
+	"github.com/netcracker/qubership-dbaas/dbaas-operator/internal/requestcontext"
 )
 
 // DefaultLimit is the page size requested per poll. Excess changes are drained on
@@ -107,6 +108,10 @@ func (p *RotationPoller) Start(ctx context.Context) error {
 // rotations — already synced by the startup reconcile — are not replayed. A failed
 // seed returns nil so the next tick retries it.
 func (p *RotationPoller) pollOnce(ctx context.Context, cursor *aggregatorclient.ChangeCursor, limit int) *aggregatorclient.ChangeCursor {
+	// Give each poll iteration its own correlation scope instead of reusing an ID
+	// inherited from the long-lived manager context.
+	ctx, _ = requestcontext.WithFreshRequestID(ctx)
+
 	if cursor == nil {
 		resp, err := p.Source.GetChangedSince(ctx, nil, 0)
 		if err != nil {
--- a/dbaas-operator/internal/poller/rotation_poller_test.go
+++ b/dbaas-operator/internal/poller/rotation_poller_test.go
@@ -19,11 +19,21 @@ package poller
 import (
 	"context"
 	"errors"
+	"os"
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
+	"github.com/netcracker/qubership-core-lib-go/v3/context-propagation/baseproviders/xrequestid"
+
 	aggregatorclient "github.com/netcracker/qubership-dbaas/dbaas-operator/internal/client"
+	"github.com/netcracker/qubership-dbaas/dbaas-operator/internal/requestcontext"
 )
+
+func TestMain(m *testing.M) {
+	requestcontext.RegisterProviders()
+	os.Exit(m.Run())
+}
 
 // fakeSource is a stub ChangedSource that records the cursors it was called with
 // and returns a fixed response/error.
@@ -39,6 +49,40 @@ func (f *fakeSource) GetChangedSince(_ context.Context, cursor *aggregatorclient
 		return nil, f.err
 	}
 	return f.resp, nil
+}
+
+type requestIDSource struct {
+	requestIDs []string
+}
+
+func (s *requestIDSource) GetChangedSince(ctx context.Context, _ *aggregatorclient.ChangeCursor, _ int) (*aggregatorclient.ChangedDatabasesResponse, error) {
+	requestID, err := xrequestid.Of(ctx)
+	if err != nil {
+		s.requestIDs = append(s.requestIDs, "")
+	} else {
+		s.requestIDs = append(s.requestIDs, requestID.GetRequestId())
+	}
+	return &aggregatorclient.ChangedDatabasesResponse{}, nil
+}
+
+func TestPollOnce_InitializesFreshRequestID(t *testing.T) {
+	src := &requestIDSource{}
+	p := &RotationPoller{Client: newFakeClient(), Source: src}
+
+	p.pollOnce(context.Background(), nil, DefaultLimit)
+	p.pollOnce(context.Background(), nil, DefaultLimit)
+
+	if len(src.requestIDs) != 2 {
+		t.Fatalf("request IDs = %v, want two", src.requestIDs)
+	}
+	for _, requestID := range src.requestIDs {
+		if _, err := uuid.Parse(requestID); err != nil {
+			t.Errorf("request ID %q is not a UUID: %v", requestID, err)
+		}
+	}
+	if src.requestIDs[0] == src.requestIDs[1] {
+		t.Errorf("poll iterations reused request ID %q", src.requestIDs[0])
+	}
 }
 
 func TestPollOnce_SeedFromHighWaterMark(t *testing.T) {
--- a/dbaas-operator/AGENTS.md
+++ b/dbaas-operator/AGENTS.md
@@ -80,24 +80,21 @@ github.com/netcracker/qubership-core-lib-go/v3/context-propagation/baseproviders
 ### Initialization (in `main()`)
 
 ```go
-ctxmanager.Register([]ctxmanager.ContextProvider{
-    xrequestid.XRequestIdProvider{},
-})
+requestcontext.RegisterProviders()
 ```
 
 ### Per-reconcile / per-request
 
-Every reconcile loop or incoming request handler must generate a unique request ID and
-attach it to the context:
+Every reconcile loop or incoming request handler must use the shared helper to generate
+a unique request ID and attach it to the context:
 
 ```go
-requestID := uuid.New().String()
-ctx = ctxmanager.InitContext(ctx, map[string]any{
-    "X-Request-Id": requestID,
-})
+ctx, requestID := requestcontext.WithFreshRequestID(ctx)
 ```
 
 This enables end-to-end tracing across operator → aggregator → adapter logs.
+It is also a hard precondition for aggregator calls: the client rejects a request
+when the provider is not registered or the caller did not initialize the context.
 
 ---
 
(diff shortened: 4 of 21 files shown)


kichasov (commented):
Reviewed the change — the fix is correct and well-tested. Verified against the branch:

- `XRequestIdProvider` is registered in `cmd/main.go`, so the new middleware is active at runtime.
- `initReconcileContext` already seeds the context via `ctxmanager.InitContext`, so the header carries exactly the ID stored in `status.lastRequestId` — operator and aggregator logs now correlate end to end.
- Before this PR the client sent no `X-Request-Id` at all (only `Accept` was set), so this closes a real gap.
- Calls whose context was never initialized stay safe: `GetSerializableContextData` (core-lib-go v3.9.1) silently skips absent context objects, so such requests simply go out without the header.

Minor remarks, none blocking:

1. **[P3] Hard fail on serialization errors.** `return fmt.Errorf("propagate request context: %w", err)` fails the whole aggregator call. Today this branch is practically unreachable (`xrequestid` serialization cannot fail), but a future provider with a fallible `Serialize()` would start failing data-path calls. Consider a short comment stating that the fail-fast is intentional, or switch to log-and-continue.
2. **[P3] Test consistency.** The client tests guard `ctxmanager.Register` with `sync.Once`; the poller test calls it directly. Duplicate registration is harmless, but mirroring the `sync.Once` helper in the poller package would keep the pattern uniform.
3. **[P3] Empty PR description.** Worth two lines for release notes: the header was not sent before; it now matches `status.lastRequestId`; the rotation poller stamps a fresh ID per tick.
4. **Observation, no action needed.** There is no end-to-end reconcile-to-header test — the client test builds the context by hand. Same mechanism, so this is fine; OperatorIT could eventually assert the correlation on the aggregator side.

dbaas-operator/internal/client/aggregator_client.go
```diff
@@ -113,6 +114,14 @@ func newClient(baseURL string, getToken func(ctx context.Context) (string, error
 		// so the warning is pure noise logged on every call (incl. the rotation poll loop).
 		SetDisableWarn(true).
 		OnBeforeRequest(func(_ *resty.Client, r *resty.Request) error {
+			if err := ctxhelper.AddSerializableContextData(r.Context(), func(name, value string) {
+				r.SetHeader(name, value)
+			}); err != nil {
+				// Fail fast rather than send a request with partially propagated context:
+				// outbound correlation data is part of the operator-to-aggregator contract.
```
vlsi: The justification does not match what the code does in the common case.

When the context was never initialized, `GetSerializableContextData` returns an empty map and the request goes out with no `X-Request-Id` at all — silently, no warning. So the header is mandatory in the branch that is currently unreachable, and optional in the branch that is reachable today.

Pick one of the two readings:

- correlation data is best-effort — log a warning and continue, since a tracing failure should not become a reconcile error, a `BackingOff` phase, and a Kubernetes event on a data-path call;
- it is a contract — then the absent-context case deserves at least a warning too.

As written, the comment asserts a contract the surrounding code does not enforce.

dbaas-operator/internal/poller/rotation_poller.go
```diff
@@ -107,6 +110,10 @@ func (p *RotationPoller) Start(ctx context.Context) error {
 // rotations — already synced by the startup reconcile — are not replayed. A failed
 // seed returns nil so the next tick retries it.
 func (p *RotationPoller) pollOnce(ctx context.Context, cursor *aggregatorclient.ChangeCursor, limit int) *aggregatorclient.ChangeCursor {
+	ctx = ctxmanager.InitContext(ctx, map[string]any{
+		xrequestid.X_REQUEST_ID_HEADER_NAME: uuid.New().String(),
+	})
```
vlsi: Two things about these three lines.

**The choice carries no comment.** In a file where `zeroUUID`, the keyset cursor, and the unconditional cursor advance each get a paragraph, the one line that decides "a fresh ID per poll iteration rather than one per poller lifetime" — and that silently replaces any ID inherited from the parent context — has none. One sentence covers it.

**It is the third spelling of one idiom.** `initReconcileContext` (`internal/controller/helpers.go:133`) does the same thing with its own local `xRequestID = \"X-Request-Id\"` constant (`helpers.go:45`), and `AGENTS.md` documents the pattern with a bare string literal. A shared `WithFreshRequestID(ctx) (context.Context, string)` would collapse all three, and this PR is the natural moment for it.

dbaas-operator/internal/poller/rotation_poller_test.go
```diff
@@ -41,6 +48,43 @@ func (f *fakeSource) GetChangedSince(_ context.Context, cursor *aggregatorclient
...
+	} else {
+		s.requestIDs = append(s.requestIDs, requestID.GetRequestId())
+	}
+	return &aggregatorclient.ChangedDatabasesResponse{}, nil
+}
+
+func TestPollOnce_InitializesFreshRequestID(t *testing.T) {
+	registerRequestIDProvider.Do(func() {
+		ctxmanager.Register([]ctxmanager.ContextProvider{xrequestid.XRequestIdProvider{}})
+	})
```
vlsi: `sync.Once` is not the guarantee it looks like.

It serializes `Register` against `Register`, but the hazard the library documents is `Register` against *read*: "This function in not thread safe. It is prohibited to register and read contexts in the same time." In `aggregator_client_test.go`, 46 tests call `t.Parallel()` and read that same global registry; the propagation test there is safe only because it is the one non-parallel top-level test in the file. Nothing records that invariant, so adding `t.Parallel()` to it for consistency with its neighbors introduces a data race.

`internal/controller/suite_test.go` already sets the precedent — register once in suite setup. `TestMain` (or a test-file `init`) would be race-free by construction and would remove the duplicated `sync.Once` var from both packages. Failing that, a comment on each test stating that it must not be parallel.

vlsi (commented):
Reviewed the updated branch (`2e80524`). Verified independently before commenting:

- The two unit tests and the new envtest assertion are real regression tests. Reverting the `OnBeforeRequest` propagation makes each of them fail — the controller suite goes to `201 Passed | 1 Failed` on the `status.lastRequestId` expectation.
- Every endpoint routes through the single `doRequest` -> `c.rc.R()`, so the hook covers all 12 client methods, not only the one the client test exercises. The 46 existing tests that call with `context.Background()` also prove an uninitialized context does not fail the request.
- The rotation poller was the only aggregator caller outside a reconcile, and all five `Reconcile` methods already seed the context through `initReconcileContext`. No caller is left without one.
- `make test` green including envtest, `go test -race` clean on all three touched packages, `make lint` reports only the two findings already red on `feat/operator-dev`, `go mod tidy` is a no-op.

The change also makes an already-documented contract true: `api/v1/common_types.go` and `docs/howto/DBaaS Operator.md` both tell users to correlate `status.lastRequestId` with aggregator logs, which was impossible while the header was never sent.

Remaining remarks, none blocking. Three are inline; these three have no line in the diff to anchor to.

**Removing `ctxmanager.Register` from `cmd/main.go` would not fail a single test.** Every test registers the provider itself — `suite_test.go` in `BeforeSuite`, and now a `sync.Once` in each of the two other packages. Drop the call from `main()` and the build stays green, the new envtest assertion stays green, and the header silently disappears in production, which is the exact failure this PR fixes. Extracting `registerContextProviders()` from `main()` and asserting on it in a `cmd` test (that package already has tests) would close the loop.

**The trace still breaks at the fan-out.** The poller's request ID covers `GetChangedSince` only. `PatchClaimsForRotation` stamps `AnnotationRotationTrigger` with a timestamp, and the reconcile it wakes mints its own ID, so poll -> fan-out -> secret refresh cannot be linked back to the poll that caused it. Out of scope here, but the end-to-end tracing claim in `AGENTS.md` is still only half delivered.

**Commit subject.** Conventional Commits wants the imperative: `fix: propagate request id to the aggregator`. release-drafter copies the subject into release notes verbatim.

kichasov (approved):
Re-reviewed the branch at `d3f73c6d`. Approving — every remark from my earlier review is addressed, and the strengthening commit is verified safe:

- The fail-fast intent is now both documented and actually enforced: an empty `X-Request-Id` rejects the request client-side before it is sent, with a negative test proving the request never reaches the server. This resolves the comment-vs-behavior contradiction from the earlier review round.
- I re-checked every production path into the aggregator client on this branch: all six controllers seed the context via `initReconcileContext` as their first step, the rotation poller stamps a fresh UUID per tick, and `cmd/` only constructs the client. No caller is left without a seeded context, so the enforcement cannot misfire today.
- The envtest now asserts the captured header equals `status.lastRequestId` — the end-to-end reconcile-to-header proof.
- Both test packages register the provider in `TestMain`, which is race-free by construction against the parallel tests reading the registry.
- The PR description is filled; one nit: it mentions reconcile requests only, while the fix also covers the rotation poller.

Status of the remaining open remarks from the other review (none blocking, agreed):

1. **Nothing pins `ctxmanager.Register` in `main.go`** — still true; note the enforcement commit changed the failure mode from a silently missing header to every aggregator call failing loudly, so a regression would be caught immediately in any live run, though not by unit CI. Extracting `registerContextProviders()` with a `cmd` test is cheap and worth a follow-up.
2. **Three spellings of the fresh-request-id idiom** (`helpers.go` local constant, inline `InitContext` in the poller, the AGENTS.md literal) — a shared `WithFreshRequestID(ctx)` helper would collapse them; follow-up material.
3. **Trace break at the rotation fan-out** (the poll ID does not reach the claim reconciles triggered via the annotation) — out of scope here as already noted; deserves an issue. Carrying the poll's request ID in the annotation value is one possible design.

Recommending these three be tracked as a follow-up PR (1–2) and an issue (3) so they don't get lost.

dbaas-operator/internal/client/aggregator_client.go
TheHollowMonarch: Added a check to ensure the comment matches the code

kichasov: Verified on the current head: this is now resolved by the "contract" reading. Commit 63cfeac9 (with d3f73c6d) makes the middleware reject any request whose X-Request-Id is empty after propagation — before it is sent, with a negative test proving the request never reaches the server — and the reworded comment now captions both guards accurately. I also re-checked that every production caller seeds the context (all six controllers via initReconcileContext, the poller per tick; cmd/ only constructs the client), so the enforcement cannot misfire today.

dbaas-operator/internal/client/aggregator_client.go
```diff
@@ -113,6 +115,18 @@ func newClient(baseURL string, getToken func(ctx context.Context) (string, error
...
+				r.SetHeader(name, value)
+			}); err != nil {
+				// Fail fast when correlation context cannot be propagated;
+				// X-Request-Id is required by the operator-to-aggregator contract.
+				return fmt.Errorf("propagate request context: %w", err)
+			}
+			if r.Header.Get(xrequestid.X_REQUEST_ID_HEADER_NAME) == "" {
+				return fmt.Errorf("propagate request context: required header %s is missing",
+					xrequestid.X_REQUEST_ID_HEADER_NAME)
+			}
```
vlsi, line 131: Three remarks about the guard, all about how the failure surfaces rather than whether it should exist.

**It is classified as a network error.** The error leaves `OnBeforeRequest` as a plain `error`, not an `*AggregatorError`, so `handleAggregatorError` (`internal/controller/helpers.go:279`) takes the "5xx / network - transient" branch and `aggregatorResult` (`internal/controller/metrics.go:180`) labels the metric `network_error`. A permanent wiring mistake then retries forever under backoff and reads as a network problem on the dashboard. `Stalled=True` / `InvalidConfiguration` matches what actually happened.

**The repo already answers this question differently.** `requestIDFromContext` (`internal/controller/helpers.go:123`) hits the same invariant — context not initialized — and logs, then panics with an explanatory message. The guard here answers it with a quiet transport error on every request, as far from the cause as the code allows. Two conventions for one invariant.

**It checks the header, not the context.** `r.Header.Get(...) == \"\"` collapses two distinct causes into one message: the provider was never registered, and the caller never seeded the context. `xrequestid.Of(ctx)` separates them. The current text also names the symptom without an action — "required header X-Request-Id is missing" leaves the reader to work out that a caller upstream has to seed the context.

vlsi (commented):
Re-reviewed at `a460f50`. The two threads I opened that are fully addressed are resolved; the one on `rotation_poller.go` stays open because only half of it landed (the comment is in, the duplicated idiom is not).

Verified on this head before commenting:

- `make test` green (`client` 92.9%, `controller` 80.4%), `go test -race ./internal/...` clean, `make lint` reports only the two findings already red on `feat/operator-dev`.
- All eight reconcile entry points seed the context — the five `Reconcile` methods plus `ReconcileMicroservice` / `ReconcileNamespace` / `ReconcilePermanent` in `balancingrule_controller.go`. The `enqueue*` map functions do not seed, but they never touch the aggregator. So the new guard cannot misfire today; that part of @kichasov's check holds.
- Mutation check on the new failure mode: dropping the provider registration from the `client` package `TestMain` fails **59 of 89 tests**. That is the shape of the production outage described below.

One point worth settling before merge, and three smaller ones. Guard-specific remarks are inline.

**The enforcement commit repriced the `main.go` follow-up.** Until now, a missing `ctxmanager.Register` in `main()` cost a missing header. It now fails every aggregator call: all CRs land in `BackingOff` and retry forever. The failure is loud, but "loud" here means the operator does nothing at all, and unit CI still would not catch it — every test registers the provider itself. Extracting `registerContextProviders()` and asserting on it in a `cmd` test moved from a nice-to-have to the thing that pins the new invariant. Worth doing in this PR rather than deferring it.

**`AGENTS.md` has not caught up.** The "Context Propagation & Request IDs" section still presents seeding as a tracing aid. It is now a hard precondition for any aggregator call: a future Runnable that forgets `initReconcileContext` gets a dead call path, not a missing log field. One line in that section covers it, and the file is the one contributors and agents are told to follow.

**Branch history.** `d3f73c6d` and `63cfeac9` are two variants of the same change under the same subject, joined by a merge of the branch into itself (`a460f50`). They differ only in two lines of comment wording, and the merge kept `63cfeac9`. Harmless if this repo squashes on merge; otherwise worth flattening, and note the approval sits on `d3f73c6d`, whose wording is not what shipped.

**Nit already raised and still open:** the PR description mentions reconcile requests only, while the fix also covers the rotation poller.

TheHollowMonarch:
Implemented the following:
- Startup context-provider registration is now extracted into registerContextProviders() and pinned by a cmd regression test.
- Fresh request-ID creation is centralized in requestcontext.WithFreshRequestID(ctx) and used by controllers, the rotation poller, tests, and AGENTS.md.
- The poller now clearly documents why each iteration creates a new correlation scope.
- Request-context validation now distinguishes missing provider registration, an uninitialized caller context, and failed context serialization with actionable messages.
- Controller request-context failures now use a typed error and produce permanent InvalidConfiguration / Stalled=True outcomes without retry, while controller metrics report configuration_error instead of network_error.
- The competing panic-based request-ID lookup was removed, and both the shared controller path and the separate InternalDatabase polling path have regression coverage.
-The context contract, metrics documentation, and dashboard were updated consistently.

The following remains for a separate follow-up:
- Poller-to-triggered-reconcile causal linking requires a parent-correlation design across the asynchronous Kubernetes work queue.

PR #649, opened 2026-08-12
feat: isolate controller rate limiters

TheHollowMonarch:
## Summary
- Giving each controller an independent composite rate limiter, preventing cross-kind backoff interference for resources sharing the same namespace and name.


kichasov (commented):
Reviewed the branch (multi-angle pass, findings cross-verified against client-go v0.36.3 / controller-runtime v0.24.1 sources). The core fix is sound: all 8 controllers (5 single + 3 balancing-rule) get isolated fresh composite limiters, `wait.Jitter` is thread-safe, `go vet` is clean, the new tests are stable over 30 consecutive runs, and repo conventions (license headers, gofmt, log style) are all observed. The remarks below are about unfinished edges, not about the fix itself.

**1. [P2] The token bucket is untunable and silently breaks the documented cap.** `rate=10`/`burst=100` are compile-time constants while base/max delays are flags. `TypedBucketRateLimiter.When` reserves a token on every `MaxOf` call and never refunds it, so reservations queue without bound: with >3000 failing CRs at the 5m cap (>10 retries/s on one controller), retry delays grow **linearly past the documented `--backoff-max-delay` cap**, and no flag or Deployment-args edit can raise the limit — contradicting the Startup Flags doc. Side effect: a NEW object's first retry is starved behind old items' retries (~(N−100)/10 s instead of the table's "1s to 1.1s"). Note that controller-runtime v0.24.1's own default for the priority queue (on by default) deliberately omits the bucket. Suggested: add `--backoff-rate/--backoff-burst` flags (or explicitly document the fixed values) and correct the delay table for the bucket-bound regime.

**2. [P2] Jitter is disabled exactly where the docs promise it helps.** `delay >= r.maxDelay` returns the cap unjittered: after ~9 failures with defaults every object sits at exactly 300s, so during an aggregator outage longer than ~9 minutes hundreds of CRs converge to synchronized retry waves with frozen phases — while the Reconcile Backoff section states jitter "spreads retries after a shared dependency outage". Confirmed independently by four analysis angles. Suggested: jitter downward within `[0.9*cap, cap]`, or document that spreading stops at the cap.

**3. [P2] The fresh-limiter-per-controller invariant is enforced only by call-site discipline.** Five reconcilers take `Options` by value while BalancingRule takes a `func() Options` factory, and no test asserts that the three balancing-rule controllers actually get independent limiters. A natural-looking refactor — hoisting `opts := optionsForController()` above the three registrations — reintroduces exactly the shared-limiter bug this PR fixes, with green CI. Suggested: move the option constructor into `internal/controller` so every `SetupWithManager` builds its own options (sharing becomes unrepresentable), or add a limiter-identity assertion over the registered controllers.

**4. [P3] `TestControllerOptionsIncludeTokenBucket` is wall-clock flaky.** The 101-call loop asserts the 101st delay > 1ns against a real-clock `rate.Limiter`: any ≥100ms stall during the loop (GC pause, CPU-starved CI runner) refills a token and the final delay becomes 0, tripping `t.Fatalf` with zero margin. Suggested: overshoot to 2×burst and assert the final delay > 1s, or build the bucket with an injectable clock.

**5. [P3] Dead `SetupWithManagerOpts` on NamespaceBinding bypasses the design.** Zero callers repo-wide; it registers with empty `Options`, which under controller-runtime v0.24.1 defaults resolves to the stock 5ms–1000s limiter — no flags, no bucket, no jitter. Delete it or route it through the shared constructor.

**6. [P3] Re-implemented client-go pieces.** The `maxDelay` field plus capping logic duplicate `workqueue.NewTypedWithMaxWaitRateLimiter`; the test's `fixedRateLimiter` duplicates `NewTypedItemFastSlowRateLimiter` with equal delays; the injectable jitter func exists only so one test can stub it (testing the `[delay, min(1.1*delay, cap)]` bounds over ~100 draws removes the field). The `[T comparable]` generic is instantiated once in production — a concrete type over `reconcile.Request` would do.

**7. [P3] Doc/test drift.** The delay table reads as upper bounds but the bucket can exceed them; the isolation invariant is stated twice within 30 lines of the same section; and the "living spec for the BackingOff retry behavior" comment repeated in four controller test files now describes a bare exponential limiter, while production runs the jittered+bucketed composite.

Verdict: the fix itself is correct and well-tested; 1–3 are worth closing in this PR or an immediate follow-up (1 and 2 concern precisely the mass-failure scenario the limiters exist for), 4 is a cheap flake prevention.

kichasov (approved):
Re-reviewed at `b6fab987` (full multi-angle re-pass, cross-checked against controller-runtime v0.24.1 / client-go v0.36.3 sources). Approving — every remark from my earlier review is addressed, several better than asked:

- **Token bucket** — resolved by removal: the bucket is gone (matching controller-runtime's own priority-queue default, which omits it), `golang.org/x/time` demoted to indirect, docs rewritten for exponential-only. The untunable constants, the linear delay growth past the cap, and the new-object starvation all disappear with it.
- **Jitter at the cap** — fixed exactly as suggested: downward jitter into `[0.9·cap, cap]` with an explanatory comment, a dedicated test (`SpreadsRetriesAtMaximumDelay`), and the docs table updated to "N-th (at cap) | 4m30s to 5m".
- **Unprotected isolation invariant** — fixed structurally: the constructor moved into `internal/controller` as `RateLimiterConfig` (immutable value, unexported fields), all six `SetupWithManager` signatures take it uniformly and build fresh options inside, and the three balancing-rule registrations each call `config.controllerOptions()`. Sharing a limiter from `main.go` is now unrepresentable, and `TestRateLimiterConfigCreatesIndependentBackoffState` asserts the independence directly.
- The flaky bucket test is gone with the bucket; dead `SetupWithManagerOpts` deleted; the generic dropped for a concrete `reconcile.Request` type; the injectable jitter stub replaced by bounds assertions over the upstream `NewTypedItemFastSlowRateLimiter`; the duplicated doc sentence and the stale "living spec" comments removed. Keeping the hand-rolled cap is now justified — it carries the downward-jitter logic that `NewTypedWithMaxWaitRateLimiter` cannot express.

I also verified the wrapper contract against the default priority queue (one `When` per rate-limited add, `Forget`/`NumRequeues` delegated, `after <= 0` semantics preserved) and that all 8 registrations route through `controllerOptions()`.

Non-blocking follow-ups the re-pass surfaced — worth a small cleanup PR:

1. **Validate in `NewRateLimiterConfig`** (base > 0, max >= base). `--backoff-base-delay=0` (accepted by `flag.DurationVar`) or a future `RateLimiterConfig{}` yields a zero-delay hot loop via the `delay <= 0` branch — whereas the old nil-`RateLimiter` path fell back to controller-runtime's safe defaults. The PR created the single chokepoint where a clamp is one line.
2. **Sync the guidance**: `AGENTS.md:274` still says to use `workqueue.NewTypedItemExponentialFailureRateLimiter` directly and the `SetupWithManager` skeleton (~line 511) still shows `opts ctrlcontroller.Options` — a new controller scaffolded from the doc reintroduces the no-jitter pattern this PR removes. `AGENTS.md:56` also quotes the deleted lowercase log line, and the `--backoff-*` `--help` text still describes exact doubling with no mention of jitter.
3. **Derive the logged jitter from the constant**: `main.go` prints a hardcoded `"10%"` while `backoffJitterFactor` lives unexported in the controller package — export it or add a `String()`/accessor on `RateLimiterConfig` so the log cannot drift.
4. **Two regression-guard gaps**: no test binds the `min(...)` clamp through the production path (deleting it keeps the suite green), and dropping `.WithOptions(config.controllerOptions())` from any controller fails no test.
5. Cosmetic: naming the jitter slice in the at-cap expression (`jitter := wait.Jitter(max, f) - max; return max - jitter`) would make the `[0.9·cap, cap]` math readable at a glance.

ArkuNC (approved):

